PR #1, opened 2018-11-14
Model release stages and support basic git scenario

joeyAghion:
This introduces the basic modeling of organizations, projects, stages, snapshots, and comparisons. The hope is for this project to present a practically real-time view of pending work in the release pipeline (e.g., merged but not built, or in staging but not production).

The basic case of shelling out to `git` works via [`releasecop`](https://github.com/joeyAghion/releasecop), which could be scheduled as a cron. An unformatted view lists the current state. `hokusai`-type release stages are still unsupported.

Models:
* `Organization` e.g., "artsy"
  * `Organization.has_many :projects`
* `Project` e.g., "helix"
  * `Project.has_many :stages`
  * `Project.has_many :snapshots`
* `Stage` e.g., "master," "staging," or "production"
* `Snapshot` a point-in-time state of all of a project's stages
  * `Snapshot.has_many :comparisons`
* `Comparison` a diff between 2 consecutive project stages (possibly empty)


joeyAghion:
The only interesting logic is in `ComparisonService`.

Issue #8, opened 2018-11-20
Design mockup


orta:
So I can find it later

![design](https://user-images.githubusercontent.com/49038/48809334-2b730880-ecf2-11e8-9d06-1741c8b488f6.png)

PR #9, opened 2018-12-04
Adds a view mode for putting on Artsy's dashboard screens

orta:
Creates the UI in #8 

There's an example using staging data in [slack](https://artsy.slack.com/archives/C2MEQ2HA6/p1543943755059300)

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,9 +1,18 @@
+# frozen_string_literal: true
+
 class ProjectsController < ApplicationController
   def index
     @organization_id = projects_params[:organization_id]
     @projects = Project.where(projects_params).sort_by do |project|
       [project.fully_released? ? 1 : 0, project.name]
     end
+  end
+
+  def simple
+    @organization_id = params[:organization_id]
+    @projects = Project.where(projects_params).to_a
+    @released = @projects.select(&:fully_released?).sort_by(&:name)
+    @needs_work = @projects.reject(&:fully_released?).sort_by { |p| p.snapshots.length }
   end
 
   private
--- a/app/views/organizations/index.html.erb
+++ b/app/views/organizations/index.html.erb
@@ -1,5 +1,8 @@
 <ul>
   <% @organizations.each do |org| %>
-    <li><%= link_to org.name, projects_path(organization_id: org.id) %></li>
+    <li>
+        <%= link_to org.name, projects_path(organization_id: org.id) %>, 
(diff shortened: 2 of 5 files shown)


app/controllers/projects_controller.rb
```diff
@@ -6,6 +8,13 @@ def index
     end
   end
 
+  def simple
+    @organization_id = params[:organization_id]
+    @projects = Project.where(projects_params).to_a
+    @released = @projects.select(&:fully_released?).sort_by(&:name)
+    @needs_work = @projects.reject(&:fully_released?).sort_by { |p| p.snapshots.length }
```
joeyAghion: Old snapshots are automatically cleaned up [here](https://github.com/joeyAghion/horizon/blob/2d2069ce49c1091935c6240cdf329c002914e0cb/app/services/comparison_service.rb#L66-L70), so most projects will have the same number. Maybe alphabetical still makes sense for this list?

app/views/organizations/index.html.erb
```diff
@@ -1,5 +1,8 @@
 <ul>
   <% @organizations.each do |org| %>
-    <li><%= link_to org.name, projects_path(organization_id: org.id) %></li>
+    <li>
+        <%= link_to org.name, projects_path(organization_id: org.id) %>, 
+        <%= link_to "dashboard", "/projects/" + org.id.to_s + "/dashboard" %>
```
joeyAghion: I'd like to structure this URL differently... maybe just make it an alternate param to the `index` action. (Happy to mess with it after merging... just don't set this path in stone anywhere in the meantime.)

config/routes.rb
```diff
@@ -3,4 +3,5 @@
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
   root to: 'organizations#index'
   resources :projects, only: [:index]
+  get 'projects/:organization_id/dashboard', to: 'projects#simple'
```
joeyAghion: The other view is more "simple!" We should just call this "dashboard" consistently. (This can wait on the other path changes I mentioned.)

app/controllers/projects_controller.rb
orta: I'm actually interested in the number of commits, so this is fundamentally wrong :D

app/views/organizations/index.html.erb
orta: no worries to me

joeyAghion:
As discussed, I'll merge and make some of the suggested modifications.

PR #10, opened 2018-12-10
Automatically refresh upon new snapshot data

joeyAghion:
I keep a tab open with the list of projects, and apparently a view is up on the wall screens as well, but until now they wouldn't refresh to reflect changes. This uses `actioncable` to broadcast a notification whenever there are changes to project comparisons. The index view(s) reload themselves based on this notification.

A cron is responsible for refreshing project comparisons currently. I refactored it slightly to iterate through projects on a per-organization basis and broadcast a notification after processing each organization, and optionally again after all organizations (for any viewers who are not filtering by `organization_id`).

This is my first use of `actioncable` so I wasn't sure how best to share server data with the client code and settled on just putting the filter params into a hidden field for the client-side code to pick up.

This works locally but will probably need some special configuration in kubernetes to support websockets. Will look at that next.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ gem 'jbuilder', '~> 2.5'
 gem 'acts_as_list' # order Project#stages
 gem 'releasecop', '>= 0.0.13' # compare release stages
 gem 'activeadmin' # manage models
+gem 'redis' # actioncable adapter
 
 group :development, :test do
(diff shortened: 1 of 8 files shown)


joeyAghion:
I'll merge this prior to figuring out the kubernetes modifications since I don't think it will interfere with anything.

PR #11, opened 2018-12-10
Revise the UI to fit the artsy design system

orta:
Pic in [slack](https://artsy.slack.com/archives/CC56D129X/p1544466467186700) but this should wrap up the design feedback and well as product feedback, now that we've got something to show up on 25. 

I've already migrated the db on prod, so that I could work with the description in prod.

### DIFF
--- a/app/views/projects/_index_dashboard.html.erb
+++ b/app/views/projects/_index_dashboard.html.erb
@@ -2,7 +2,8 @@
 <link rel="stylesheet" type="text/css" href="https://webfonts.artsy.net/unica-webfonts.css" />
 <style type="text/css">
 body {
-  font-family: "Unica77LLWebRegular"
+  font-family: "Unica77LLWebRegular";
+  font-size: 14px;
 }
 
 h1 {
@@ -12,11 +13,14 @@ h1 {
   line-height: 24px;
   clear: both;
   margin: 30px;
+  font-family: "Unica77LLWebMedium";
+  font-weight: normal;
 }
 
 h2 {
   font-size: 20px;
-  font-weight: bold;
+  font-family: "Unica77LLWebMedium";
+  font-weight: normal;
   color: black;
 }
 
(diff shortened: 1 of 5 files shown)


app/views/projects/_index_dashboard.html.erb
```diff
@@ -2,144 +2,200 @@
...
+  font-size: 28px;
+  margin: 0;
+  padding: 0;
+}
+
+.todo .project .comparison {
+  margin-bottom: 0px;
   margin-right: 16px;
+  min-height: 40px;
+  background: url(/assets/arrow-d9bccef77260cf5af9399ae03a67b3e0f78b0ab67a81224e6dab827ba2ffbe4c.svg) 0.5px 20px no-repeat;
```
joeyAghion: This should use the built-in helpers rather than hard-code an asset digest. In this context I think this would work:

    background: url(<%= image_url('arrow.svg') %>) 0.5px 20px no-repeat;

But better would be to move these to `projects.scss` where they can use things like `image-url(...)` inline.

joeyAghion:
Almost forgot to ask: what do the red/yellow/green colors mean? I'm not sure I see the pattern.

app/views/projects/_index_dashboard.html.erb
```diff
@@ -2,144 +2,200 @@
...
             <div class="comparison">
-              <div class="number"><%= link_to comparison.description.length, params.permit(:organization_id).merge(view: 'detail', anchor: project_anchor(project)), 'data-turbolinks' => false %></div>
-              <div class="release">
-                <%= ahead.name.titleize %> to<br /><%= behind.name.titleize %><br />
-              </div>
-              <div class="who">
-                <%= comparison.description.map { |line| first_name_from_log(line).titleize }.uniq.to_sentence %>
-              </div>
+              <% commit_length = comparison.description.length %>
+              <% circle_color = commit_length == 0 ? "green" : commit_length > 10 ? "red": "yellow" %>
```
joeyAghion: Now that I'm looking more closely, I'm not sure I agree with this classification... it's more about the stage (e.g., failing a master build) or age (e.g., undeployed for X days) than the number of commits.

Either way we can refine this later. (When we do, this could be a helper rather than repeat the logic throughout the template.)

PR #30, opened 2019-05-31
@ashkan18: Allow projects to be tagged and allow filtering the dashboard to those tags

joeyAghion:
Once I started investigating how to make the sorting more sophisticated, I realized how it would require completely replacing `releasecop` with a more structured, programmatic alternative.

This to-do about adding tags to projects seemed much easier.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -3,7 +3,9 @@ class ProjectsController < ApplicationController
 
   def index
     @organization_id = projects_params[:organization_id]
-    @projects = Project.where(projects_params).sort_by do |project|
+    @projects = Project.where(projects_params)
+    @projects = @projects.where('tags ?| array[:tags]', tags: params[:tags]) if params[:tags]&.any?
+    @projects.sort_by do |project|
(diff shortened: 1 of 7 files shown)


app/models/project.rb
```diff
@@ -1,9 +1,13 @@
 class Project < ApplicationRecord
+  include JsonbEditable
```
ashkan18: 😍

app/controllers/projects_controller.rb
```diff
@@ -3,7 +3,9 @@ class ProjectsController < ApplicationController
 
   def index
     @organization_id = projects_params[:organization_id]
-    @projects = Project.where(projects_params).sort_by do |project|
+    @projects = Project.where(projects_params)
+    @projects = @projects.where('tags ?| array[:tags]', tags: params[:tags]) if params[:tags]&.any?
```
ashkan18: ❤️ Postgres

joeyAghion: Yeah this was a little too easy... I suspect problems may turn up when I try it in environments with more than my one test project.

PR #34, opened 2019-08-09
Minor refactors

joeyAghion:
* Combined some test stubs.
* `ComparisonService` was getting big, so I moved all the helpers that were specific to interacting with `releasecop` and parsing its output into a dedicated service class. This also became useful in the view helpers.
* Adjusted the deploy threshold to `> 10` commits to match the dashboard treatment.


joeyAghion:
Gonna merge to submit an additional PR with actual logic changes on top, but feedback still welcome.

PR #35, opened 2019-08-09
@mzikherman: Determine deploy severity based on commits, contributors, and age

joeyAghion:
Took a stab at a slightly more sophisticated way of determining deploy "severity."

This penalizes based on:
* number of commits
* number of distinct contributors
* age of oldest work, in days

To do (later):
* [] Update dashboard treatment to match new logic.
* [] Consider parameterizing `releasecop`'s behavior so we can see merge commits in its output. This would allow us to count PRs instead of commits, which would be slightly more meaningful.

I stuck with our existing threshold of `10` for deciding a diff warrants a deploy. This means:
* one commit by a single contributor will qualify after ~3 days
* one commit each by 2 contributors will qualify after ~2 days
* one commit each by 3 contributors will qualify immediately

However the "age" isn't a perfect measure since it corresponds with when the commit was authored, not when it was merged. We might want to increase the threshold to account for that. See the chart for alternatives:

### DIFF
--- a/app/services/comparison_service.rb
+++ b/app/services/comparison_service.rb
@@ -2,6 +2,7 @@ class ComparisonService
   attr_accessor :project
 
   KEEP_OLD_SNAPSHOTS = 5
+  DEPLOY_AT_SEVERITY = 10
 
   def initialize(project)
     @project = project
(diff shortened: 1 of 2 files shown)


joeyAghion:
I've queued up the additional changes to consolidate the new severity metric with the visuals here, but I'll wait to PR it until after this has been deployed successfully: https://github.com/joeyAghion/horizon/pull/1

mzikherman:
I like it! I agree with the criteria (num. commits, age, and num. contributors). Can always tweak later.

mzikherman:
(Gah sorry forgot to merge last night).

PR #37, opened 2019-08-13
Refactor view logic and memo-ization into ProjectsPresenter

joeyAghion:
## Problem

The `projects#index` action is responsible for rendering both the detailed and dashboard views. It was already somewhat complicated, with some shared logic for sourcing relevant projects and lots of data-massaging and computation happening inline in the views. The recent work in https://github.com/artsy/horizon/pull/35 introduces a more sophisticated calculation of how badly a project needs a deploy, and it would be best for the views to use that same logic for rendering projects (e.g. "red"), but this would introduce even more complexity.

## Solution

Introduce a presenter class that can encapsulate some of the shared logic and provide dead-simple methods for the views to call. It can be tested more straightforwardly and can memo-ize expensive computations. An embedded wrapper class for bare `Project` models helps keeps logic out of the model classes. Once refactored, it's easy to call the existing `ComparisonService.severity_score` [that's used to decide when to auto-deploy] and make it available to the views for each project.

### DIFF
--- a/app/presenters/projects_presenter.rb
+++ b/app/presenters/projects_presenter.rb
@@ -1,0 +1,62 @@
+class ProjectsPresenter
+  VIEWS = %w[detail dashboard]
+
+  def initialize(params)
+    @params = params
+  end
+
+  def detailed_projects
+    projects.sort_by { |p| [p.fully_released? ? 1 : 0, p.name] }
+  end
+
+  def unreleased_projects
+    projects.reject(&:fully_released?).sort_by(&:severity).reverse
+  end
+
+  def released_projects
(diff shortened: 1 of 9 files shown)


app/presenters/projects_presenter.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def released_projects
+    projects.select(&:fully_released?).sort_by(&:name)
+  end
+
+  def view
+    VIEWS.include?(@params[:view]) ? @params[:view] : 'detail'
+  end
+
+  class ProjectWrapper
```
ashkan18: oooh interesting pattern! never used a class within a class.

PR #39, opened 2019-08-13
Assign appropriate user to release PRs

joeyAghion:
After creating a new release PR, this loads the associated commits and attempts to identify the most natural assignee. It considers `author`s and `committer`s (since they can be different) and tries to exclude bots, users suspected of being bots, and the special [web-flow](https://github.com/web-flow) github user.

(Not much to it!)

### DIFF
--- a/app/services/deploy_service.rb
+++ b/app/services/deploy_service.rb
@@ -10,15 +10,24 @@ class DeployService
     access_token = deploy_strategy.profile&.basic_password
     raise 'A profile and basic_password are required for Github authentication' if access_token.blank?
 
+    repo = deploy_strategy.github_repo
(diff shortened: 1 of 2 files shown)


app/services/deploy_service.rb
```diff
@@ -10,15 +10,24 @@ def self.create_github_pull_request(deploy_strategy)
...
+      pr = client.create_pull_request(
+        repo,
         deploy_strategy.arguments['base'],
         deploy_strategy.arguments['head'],
         'Deploy',
         'This is an automatically generated release PR!'
       )
+
+      # assign appropriate github user
+      sorted_logins = client.pull_request_commits(repo, pr.number).
```
zephraph: This is _very_ clever. I love it! 🙌

PR #40, opened 2019-08-16
Respect blocks when automatically starting deploys

joeyAghion:
Our new concept of deploy strategies should respect our new concept of deploy blocks.

### DIFF
--- a/app/services/comparison_service.rb
+++ b/app/services/comparison_service.rb
@@ -54,9 +54,11 @@ class ComparisonService
       new_snapshot = store_new_snapshot!(project, result, refreshed_at)
       clean_up_old_snapshots
     end
-    project.stages.select { |s| warrants_deploy?(s) }.each do |stage|
-      stage.deploy_strategies.each do |strategy|
-        DeployService.start(strategy) if strategy.automatic?
+    unless project.deploy_blocks.unresolved.any?
+      project.stages.select { |s| warrants_deploy?(s) }.each do |stage|
+        stage.deploy_strategies.each do |strategy|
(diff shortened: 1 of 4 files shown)


joeyAghion:
Just added a commit that reorganized specs _only_. The small behavior change is reviewable in the first commit.

app/services/comparison_service.rb
```diff
@@ -54,9 +54,11 @@ def refresh_comparisons
       new_snapshot = store_new_snapshot!(project, result, refreshed_at)
       clean_up_old_snapshots
     end
-    project.stages.select { |s| warrants_deploy?(s) }.each do |stage|
-      stage.deploy_strategies.each do |strategy|
-        DeployService.start(strategy) if strategy.automatic?
+    unless project.deploy_blocks.unresolved.any?
```
dleve123: slight preference for `if project.deploy_blocks.unresolved.empty?` to keep things more positive, bit easier to understand IMO.

spec/features/organizations_spec.rb
```diff
@@ -123,5 +122,24 @@
       expect(DeployService).not_to receive(:start)
       ComparisonService.new(project).refresh_comparisons
     end
+
+    it 'respects deploy blocks' do
```
dleve123: thoughts on a spec description that better deblurs what "respects" means? How about "does not kick off a deploy if there are any unresolved deploy blocks"?

app/services/comparison_service.rb
joeyAghion: I actually had this as `if ... unresolved.none?` initially but the double negative felt bad. I think I'll go with your suggestion.

PR #41, opened 2019-08-16
Display when a project's deploys are blocked

joeyAghion:
`DeployBlock`s were added in https://github.com/artsy/horizon/pull/32. This adds a visual indicator to the dashboard when a project's deploys are blocked. Hovering on the blocked symbol shows the block description(s).

Looks like:

### DIFF
--- a/app/helpers/projects_helper.rb
+++ b/app/helpers/projects_helper.rb
@@ -55,4 +55,10 @@ module ProjectsHelper
     else 'red'
     end
   end
+
+  def blocked(project)
+    blocks = project.deploy_blocks.unresolved.to_a
+    return if blocks.empty?
+    tag.div(raw('&#9888;'), class: 'blocked', title: blocks.map(&:description).to_sentence)
+  end
 end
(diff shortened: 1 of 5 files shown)


app/helpers/projects_helper.rb
```diff
@@ -55,4 +55,10 @@ def healthy_count_class(count)
     else 'red'
     end
   end
+
+  def blocked(project)
+    blocks = project.deploy_blocks.unresolved.to_a
```
dblandin: Do we expect that a project might have more than one unresolved deploy block?

What do you think about limiting projects to at most one unresolved deploy block?

As a client initiating a block via the API, I feel like you should get a `422 Unprocessable Entity: A deploy block already exists for this project: <description>` error if one already exists.

```ruby
if (deploy_block = project.unresolved_block)
  tag.div(raw('&#9888;'), class: 'blocked', title: block.description)
end
```

cc/ @dleve123 @zephraph

joeyAghion: I dunno I think the initial modeling continues to make sense. If I create a block because X needs QA and you record another because of a critical bug discovered in staging, they're distinct and _both_ should be resolved before the project is cleared for release.

(I do agree that "whether a project is blocked at all" is an important higher level attribute of the project though.)

dblandin: Gotcha! Makes sense :+1:

PR #44, opened 2019-08-19
Fix link helper and add test that renders a few admin views

joeyAghion:
This fixes a link helper call.

It might be controversial to test activeadmin views but at this point we've mistakenly broken those views a few time so I thought at least a lightweight test was warranted.


dleve123:
> It might be controversial to test activeadmin views but at this point we've mistakenly broken those views a few time so I thought at least a lightweight test was warranted.

I think it's controversial to *not* test for anything non-trivial. Thank you for adding!

PR #52, opened 2019-08-30
Broadcast over websockets when deploy blocks change

joeyAghion:
I noticed that dashboards don't necessarily refresh when deploy blocks change. This fixes that.

### DIFF
--- a/app/models/deploy_block.rb
+++ b/app/models/deploy_block.rb
@@ -3,4 +3,14 @@ class DeployBlock < ApplicationRecord
 
   scope :unresolved, -> { where(resolved_at: nil) }
   scope :resolved, -> { where.not(resolved_at: nil) }
+
+  after_save :broadcast_updates
+
+  private
+
+  def broadcast_updates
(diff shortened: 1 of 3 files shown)


app/models/deploy_block.rb
```diff
@@ -3,4 +3,14 @@ class DeployBlock < ApplicationRecord
...
   scope :resolved, -> { where.not(resolved_at: nil) }
+
+  after_save :broadcast_updates
+
+  private
+
+  def broadcast_updates
+    return unless id_previously_changed? || resolved_at_previously_changed?
+    ActionCable.server.broadcast(ProjectChannel.channel_name(project.organization_id), updatedBlocks: true)
+    ActionCable.server.broadcast(ProjectChannel.channel_name, updatedBlocks: true)
```
joeyAghion: This repetition is because it's possible to be viewing an org-scoped dashboard _or_ a global one. Might be worth some refactoring or redesigning to encapsulate this knowledge somewhere.

PR #54, opened 2019-09-17
Hide sensitive profile attributes

dblandin:
Configure ActiveAdmin to exclude sensitive attributes from Profile views, forms, and exports.

## Screenshots

screenshot 1568728179
screenshot 1568728188


joeyAghion:
Thanks!

app/admin/profiles.rb
```diff
@@ -14,6 +14,16 @@
 
   permit_params :name, :organization_id, :basic_username, :basic_password, :environment_input
 
+  controller do
```
dleve123: out of curiosity, why was this block needed?

dblandin: This is how you update the controller ActiveAdmin creates and manages for your model:

Here's some relevant documentation: https://github.com/activeadmin/activeadmin/blob/77507a22fdad50a295b44c2e9a76e9c8736006d6/docs/8-custom-actions.md#modifying-the-controller

We're using the mechanism here to make sure we don't overwrite `basic_password` with an empty string when updating other attributes on the object.

PR #60, opened 2020-01-06
Add aged effect for super-stale releases

joeyAghion:
A little easter egg so that the truly-stale releases stand out even more. Looks like:

Hat tip: [Trello](https://help.trello.com/article/820-card-aging)

Note: this has _not_ been approved by design, but upsetting people that these stale projects look so bad is sort of the idea.


eessex:
lol this is cute

mzikherman:
Nice!

oxaudo:
I kinda like this ancient look. Should get some projects to look like that! :)

PR #61, opened 2020-01-13
Notify #dev when a deploy PR is ready

zephraph:
Step 1 of a longer process to automate our deployments, this simply integrates slack and lets us know when a deployment PR is ready.


config/initializers/slack.rb
```diff
@@ -0,0 +1,3 @@
+Slack.configure do |config|
+  config.token = ENV['SLACK_API_TOKEN']
```
joeyAghion: This project declares [and defaults] ENV vars in [this initializer](https://github.com/artsy/horizon/blob/master/config/initializers/_config.rb) by convention, so this should be added there, then referenced here. This eases stubbing for tests and keeps available configuration visible in a single location.

More generally, though, Horizon tries to support different orgs and projects using different settings, so I wonder if it's even appropriate to have a global Slack token configured. Instead, this would make a good addition to the `Profile#environment` that each deploy strategies already reference to get the necessary authorization. Similarly, `DeployStrategy#arguments` is a json field that could offer a new [optional] `slack_channel` property indicating which channel to notify. This field is already being used above to support custom branch setups.

app/services/deploy_service.rb
```diff
@@ -28,6 +28,8 @@ def self.create_github_pull_request(deploy_strategy)
         group_by(&:login).map{|k,v| [v.size, k] }.sort.reverse.map(&:last)
       assignee = sorted_logins.detect { |l| client.check_assignee(repo, l) }
       client.add_assignees(repo, pr.number, assignee) if assignee
+      slack = Slack::Web::Client.new
+      slack.chat_postMessage(channel: '#dev', message: "<#{pr.html_url}|Deploy PR> ready for #{repo}", as_user: true)
```
joeyAghion: How should this behave if the Slack call fails? Right now I think it will cause the whole cron to exit. This project doesn't have much error-reporting in place currently, so might be better to just log a warning for now.

config/initializers/slack.rb
zephraph: If we were truly wanting to support multiple organizations, then in my mind we should likely go for the whole oauth flow and enable multi-slack organization usage. That's a bit of extra work though and we're trying to be pretty scrappy on this being as it's a side thing. 

Being as @ansor4 implemented your suggestion, maybe that's a good hold-over until we can do the full oauth setup.

app/services/deploy_service.rb
```diff
@@ -32,5 +37,12 @@ def self.create_github_pull_request(deploy_strategy)
       # PR already exists
     end    
   end
+
+  def self.post_slack_notification()
+    slack = Slack::Web::Client.new
+    slack.chat_postMessage(channel: '#dev', message: "<#{pr.html_url}|Deploy PR> ready for #{repo}", as_user: true)
+  rescue StandardError
+    # ok if we don't notify slack
+  end
```
zephraph: Just a note here:

When we enable automatic deployments, if (for whatever reason) the slack message fails to send then we should definitely _not_ queue up the automated merge, ha. That would be... unfortunate.

app/services/deploy_service.rb
```diff
@@ -32,5 +37,12 @@ def self.create_github_pull_request(deploy_strategy)
       # PR already exists
     end    
   end
+
+  def self.post_slack_notification()
+    slack = Slack::Web::Client.new
+    slack.chat_postMessage(channel: '#dev', message: "<#{pr.html_url}|Deploy PR> ready for #{repo}", as_user: true)
```
joeyAghion: Possible bug: I don't think `repo` is defined in this method's scope.

Also, I don't think we should hard-code `#dev` here. Replacing it with `deploy_strategy.arguments['slack_channel']` (also not defined in this scope 😄) would remain more generic _and_ allow different projects to notify different channels if desired.

app/services/deploy_service.rb
```diff
@@ -1,7 +1,12 @@
 class DeployService
-  def self.start(deploy_strategy)
+  def self.start(deploy_strategy, slack_token = nil)
+    Slack.configure do |config|
+      config.token = slack_token if slack_token
+    end
```
joeyAghion: I think this whole `Slack.configure` block can be removed, and the `slack_token` argument eliminated from this method. Everything else should work as-is.

Excited to see this in action!

zephraph:
@joeyAghion I made the update you requested, but the tests are failing. I _think_ the test failure is unrelated (it was failing before), but I'm not sure what to update to get it to work. It seems like some of the mocking in the tests that are failing is wrong?

joeyAghion:
The tests aren't failing on `master` so they must be related to this work. It looks like the new Slack operation adds a call to `pr.html_url`, so those 2 tests need that added to their mocks.

app/services/deploy_service.rb
```diff
@@ -28,9 +31,22 @@ def self.create_github_pull_request(deploy_strategy)
...
     rescue Octokit::UnprocessableEntity
       # PR already exists
     end    
+    pr_info
+  end
+
+  def self.post_slack_notification(deploy_strategy, pr_info)
+    slack_channel = deploy_strategy.arguments['slack_channel']
+    return unless slack_channel
+    slack_token = deploy_strategy.profile&.environment&.fetch('SLACK_API_TOKEN')
```
eessex: I see that the api token is on the `deploy_strategy` I wonder if that means we need to set a token/channel for every repo? I wonder if there are some sensible defaults we can use?

joeyAghion: The API token is actually on the `Profile` associated with the deploy strategy. `Profile`s are Horizon's model for "identity" and any associated secrets. This has been useful, e.g., because it has to connect to different git repos with different identities--sometimes even for the different stages of a single project, such as when `master` is on Github but `staging` is based on Heroku. This means we should only need to set the token in one place.

The channel setting _is_ on each deploy strategy though. This was my earlier suggestion, so that different repos can report their deploys to different channels.

eessex:
Closing in favor of #192

PR #64, opened 2020-02-19
Set POSTGRES_HOST_AUTH_METHOD to fix dev/test

eessex:
Fix dev env, plus some yml validation.

### DIFF
--- a/hokusai/development.yml
+++ b/hokusai/development.yml
@@ -7,7 +7,6 @@ services:
       - horizon-postgres
       - horizon-redis
     environment:
-      - DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true
       - DATABASE_URL=postgresql://postgres:@horizon-postgres/horizon_development
       - REDIS_URL=redis://horizon-redis
     extends:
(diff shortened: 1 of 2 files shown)


hokusai/development.yml
```diff
@@ -1,26 +1,28 @@
...
     env_file: ../.env
     depends_on:
-    - horizon-postgres
-    - horizon-redis
+      - horizon-postgres
+      - horizon-redis
     environment:
-    - DATABASE_URL=postgresql://postgres:@horizon-postgres/horizon_development
-    - REDIS_URL=redis://horizon-redis
+      - DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true
```
joeyAghion: I don't think this project uses the database_cleaner library so this shouldn't be required.

hokusai/test.yml
```diff
@@ -1,15 +1,19 @@
...
 services:
   horizon:
     command: ./hokusai/ci.sh
     depends_on:
-    - horizon-postgres
+      - horizon-postgres
     environment:
-    - DATABASE_URL=postgresql://postgres:@horizon-postgres/horizon_test
-    - RAILS_ENV=test
+      - DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true
```
joeyAghion: Ditto here

PR #68, opened 2020-02-28
Add info about required env file for setup

jpotts244:
Was getting the following errors when trying to run migrations or run server on the setup step of the readme:
```
ERROR: Couldn't find env file: /Users/jackiepotts/dev/horizon/.env
ERROR: Command 'docker-compose -f /Users/jackiepotts/dev/horizon/hokusai/development.yml -p hokusai up' returned non-zero exit status 1
```

A lot of the environment variables are defined within the config files themselves but it looks like hokusai requires an `.env` file to be present in order to run https://github.com/artsy/horizon/blob/master/hokusai/development.yml#L5

To fix the error I just added a step to copy a test env into a `.env` file. 

Let me know if there is a better way to do this! I was thinking of removing the `env_file: ../.env` from hokusai/development.yml altogether? but am not confident on the ramifications of doing so.


eessex:
Addressed in #111

PR #71, opened 2020-03-16
Upgrade to ruby 2.6

eessex:
In advance of adding react, bumps ruby base version.


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.5.1'
+ruby '2.6.0'
```
jpotts244: Jon showed me this pattern the other day, its pretty neat! Instead of a string for the version here we can read in the version from the `.ruby-version` file,  it will look something like this: 

https://github.com/artsy/diffusion/blob/master/Gemfile#L10

 and removes one thing for us to change on the next version bump 🙌

jpotts244:
Lookin good! thanks for doing this! Just one small comment

PR #72, opened 2020-03-18
[WIP] Upgrade to Rails 6.0, add webpacker & react

eessex:
DO NOT MERGE

Pair project with @jpotts244


hokusai/development.yml
```diff
@@ -2,9 +2,9 @@
 version: "2"
 services:
   horizon:
+    env_file: ../.env
```
joeyAghion, line 5: One problem we've had in other projects is that `hokusai dev` will _fail_ if this file isn't found. If you've found this useful, you might want to add a step to the set-up described in the README.

hokusai/development.yml
```diff
@@ -2,9 +2,9 @@
...
     environment:
       - RAILS_ENV=development
-      - RAILS_LOG_TO_STDOUT=true
```
joeyAghion, line 7: Why remove this? (Do the defaults change in rails 6? If so the code that references it in the environment files can be removed.)

eessex: We reference this in the `build.yml` so it shouldn't need to go into individual environments.

eessex:
Perhaps this is my own bias because I'm inexperienced with haml, but I was frustrated by the difficulty of sharing styles/components/theme in the current setup. By using Palette I hope to significantly speed up the time it takes to build out views and eliminate the need to worry about theming.

PR #87, opened 2020-04-17
Handle mix of nil and [] tags values

joeyAghion:
I noticed that production was error-ing and it ended up being due to a single project having `tags: nil`. I updated it to a [JSON] array for now, but this fixes the home page going forward.


eessex:
👍

PR #91, opened 2020-05-27
Rails 6.0 upgrade

eessex:
Opening in favor of https://github.com/artsy/horizon/pull/72, updates Rails to `6.0.3` as well as a minor Ruby jump and many gem bumps. 

This PR does not include `webpacker`/`react`, which include lots of asset pipeline changes-- will open a separate PR for this.


config/environments/production.rb
```diff
@@ -51,15 +48,15 @@
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :info
+  config.log_level = :debug
```
joeyAghion, line 51: Intentional?

eessex: Probably added by the rails bot - will change this one back

PR #95, opened 2020-06-04
Migrate asset pipeline to Webpack, use TS & React

eessex:
Adds TS/React front end toolchain:
- Installs `webpacker` gem and compile JS assets via Webpack
- Adds`typescript`, `react`, `@artsy/palette` and `styled-components`
- Adds linting/prettier and pre-commit checks
- Adds navbar component

Screen Shot 2020-06-04 at 3 21 44 PM


babel.config.js
```diff
@@ -0,0 +1,88 @@
+module.exports = function(api) {
+  var validEnv = ['development', 'test', 'production']
+  var currentEnv = api.env()
+  var isDevelopmentEnv = api.env('development')
+  var isProductionEnv = api.env('production')
+  var isTestEnv = api.env('test')
```
damassi, line 6: `var` -> `const`

config/webpacker.yml
```diff
@@ -0,0 +1,95 @@
...
+    public: localhost:3035
+    hmr: false
+    # Inline should be set to true if using HMR
+    inline: true
+    overlay: true
+    compress: true
+    disable_host_check: true
+    use_local_ip: false
+    quiet: false
+    pretty: false
```
damassi, line 71: `true`

config/webpacker.yml
```diff
@@ -0,0 +1,95 @@
...
+    port: 3035
+    public: localhost:3035
+    hmr: false
+    # Inline should be set to true if using HMR
+    inline: true
+    overlay: true
+    compress: true
+    disable_host_check: true
+    use_local_ip: false
+    quiet: false
```
damassi, line 70: `true` -- probably don't need all the of the misc chunk output

config/webpacker.yml
```diff
@@ -0,0 +1,95 @@
...
+  <<: *default
+  compile: true
+
+  # Reference: https://webpack.js.org/configuration/dev-server/
+  dev_server:
+    https: false
+    host: localhost
+    port: 3035
+    public: localhost:3035
+    hmr: false
```
damassi, line 63: `true` - it will default to full page reload if hmr for some reason can't apply updates.

package.json
```diff
@@ -1,5 +1,63 @@
...
+    "webpack-dev-server": "^3.11.0"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
+      "pre-push": "yarn type-check"
+    }
+  },
+  "lint-staged": {
+    "*.@(ts)": [
```
damassi, line 52: tsx, js, and jsx?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "module": "es6",
+    "lib": ["es6", "dom"],
+    "moduleResolution": "node",
+    "noImplicitAny": false,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": "dist",
+    "pretty": true,
+    "skipLibCheck": true,
+    "sourceMap": true,
```
damassi, line 14: i think `noEmit` can be set to `true` and then `sourceMap: false` and `outDir` ommitted. That should increase perf since everything is being handled through babel.

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "strict": true,
+    "target": "es2016",
+    "jsx": "react",
+    "types": ["node", "webpack-env", "jquery", "actioncable"]
+  },
+  "exclude": ["**/*.spec.ts", "node_modules", "public"]
```
damassi, line 20: don't want to exclude specs!

Dockerfile
```diff
@@ -24,9 +24,11 @@ RUN pip install --upgrade --no-cache-dir hokusai
...
     bundle config --global frozen 1
 
-# Set up gems
-COPY Gemfile Gemfile.lock .ruby-version ./
-RUN bundle install -j4
+# Set up gems & packages, empty cache to save space
+COPY Gemfile Gemfile.lock .ruby-version package.json yarn.lock ./
+RUN bundle install -j4 && \
+    yarn install --frozen-lockfile --quiet && \
+    yarn cache clean --force
```
joeyAghion, line 31: Splitting up the ruby dependency step from the yarn step might better utilize docker caching. E.g., a change to Gemfile should force just a re-bundling. Or (possibly better), a change to `package.json` shouldn't force gems to be reinstalled.

(Not a big deal in this project, but I'm anticipating that these will get copied elsewhere.)

app/javascript/channels/project_channel.ts
```diff
@@ -0,0 +1,23 @@
+import $ from "jquery"
+import consumer from "./consumer"
+
+consumer.subscriptions.create(
+  {
+    channel: "ProjectChannel",
+    organization_id: $("#organization_subscription_identifier").val(),
```
joeyAghion, line 7: If jquery isn't cool and you prefer a more modern/react-friendly alternative, this _might_ be its only usage.

damassi:
@jonallured - not sure what volts setup looks like, but at the min we should turn on `hmr` and `pretty`

app/javascript/channels/project_channel.ts
eessex: Oo good call - going to make a follow-up task to remove jQuery and update this logic.

eessex:
@jonallured excellent suggestion! I've added a default rake task and am running prettier there now.

package.json
```diff
@@ -1,5 +1,63 @@
 {
   "name": "horizon",
   "private": true,
-  "dependencies": {}
+  "scripts": {
+    "lint": "eslint ./app --ext ts,tsx,js,jsx --fix",
+    "prettier": "prettier --write 'app/**/*.(ts|tsx|js|jsx)'",
```
jonallured, line 6: I think you'd want an additional script like this:

https://github.com/artsy/convection/blob/master/package.json#L6

This ensures that the rake task will fail if there are prettier violations - otherwise the rake task will write those files and happily return 0!!

eessex: Will follow up to add this & rubocop

PR #106, opened 2020-06-11
Add jest

eessex:
Follow up to #95, adds jest test support.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -25,5 +25,5 @@ if Rails.env.development? || Rails.env.test?
   end
 
   Rake::Task[:default].clear
-  task default: %i[prettier spec jest]
+  task default: %i[prettier jest spec]
 end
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -18,6 +18,12 @@ if Rails.env.development? || Rails.env.test?
...
 
+  desc 'run jest'
+  task jest: :environment do
+    system 'yarn test'
+    abort 'jest failed' unless $CHILD_STATUS.exitstatus.zero?
+  end
+
   Rake::Task[:default].clear
-  task default: %i[prettier spec]
+  task default: %i[prettier spec jest]
```
jonallured, line 28: I would suggest putting jest before spec because it's so much faster, haha.

```suggestion
  task default: %i[prettier jest spec]
```

PR #111, opened 2020-06-15
Add dotenv, non-hokusai server options

eessex:
- adds `dotenv` gem to dev and test environments, for running server without hokusai/docker builds
- adds `pull_data` script to dump staging data to local db
- updates readme with instructions on server, webpack-dev-server, tests &pull_data


eessex:
There seems to be an error related to downloading `hokusai` unrelated to changes... tests are not yet running.

app/controllers/application_controller.rb
```diff
@@ -2,7 +2,7 @@ class ApplicationController < ActionController::Base
   http_basic_authenticate_with(
     name: Horizon.config[:basic_auth_user],
     password: Horizon.config[:basic_auth_pass]
-  ) if Horizon.config[:basic_auth_pass].present?
+  ) if !Rails.env.test? && Horizon.config[:basic_auth_pass].present?
```
jonallured, line 5: I was thinking about this last night and I wonder if there's a more declarative way to approach this. Like, what if this looked more like this:

```rb
http_basic_authenticate_with(Horizon.config[:basic_auth_creds]) if Horizon.config[:basic_auth_enabled]
```

And then in the test environment we set that `basic_auth_enabled` to false. That seems less mysterious to me but certainly not blocking feedback, just something I was thinking about! It also demotes this controller to knowing less about the way config happens ya know? What do you think?

.gitignore
```diff
@@ -27,6 +27,7 @@
 /config/master.key
 
 .env
+*.dump
```
jonallured, line 30: Good call!!! 😝

README.md
```diff
@@ -1,40 +1,42 @@
...
-* `Stage`s can optionally be associated with a `Profile` that stores credentials for accessing git providers or AWS.
-* `Snapshot`s capture the complete state of a project's stages at a point in time. Each `Snapshot` has associated `Comparison`s between the consecutive stages of a project (e).g., a comparison between _master_ and _staging_, and a second between _staging_ and _production_)
-* A cron periodically reevaluates these comparisons, creating a new snapshot if the state has changed at all.
-* A `Stage` (such as "production") can have a `DeployStrategy` for automatically triggering releases. Currently only the "github pull request" provider is implemented. When defined, a deploy strategy will automatically start a release (e.g., by opening a pull request) when a diff exceeds a certain threshold.
-* `DeployBlock`s indicate that a project _should not_ be released. In addition to appearing on dashboards, any unresolved blocks will prevent new releases from being automated.
-
-Setup
----
+## Setup
```
jonallured, line 19: Excellent updates to the docs!! ❤️

app/controllers/application_controller.rb
eessex: I like it! Adding now.

eessex: Decided to hold on this for the moment -- this approach requires us setting two env vars rather than one which while more declarative feels extra. I'm hoping we can convert to artsy oauth and remove this feature entirely.

jonallured: Oooo, yeah that's a better idea for sure! 🏆

PR #112, opened 2020-06-16
fix pip install

izakp:
The apk package `py-pip` was installing pip for python 3.8 - this installs pip via setuptools for python 2.7


eessex:
Thank you for this fast solution!

PR #115, opened 2020-06-16
Nav links, remove turbolinks, default to dashboard

eessex:
- Adds 'Admin' and 'Team' links to nav bar 
- Updates `MainLayout` to wrap children
- Sets up `Projects` js app and passes props from controller
- Defaults view to `projects#dashboard` (we don't need organizations with nav links)
- Removes turbolinks (causes issues loading client js)
- Add `DATABASE_URL` to test env (got a warning  about writing to production data when my .env had one already)

Screen Shot 2020-06-16 at 7 08 05 PM


.env.test
```diff
@@ -1 +1,2 @@
-BASIC_AUTH_PASS=
\ No newline at end of file
+BASIC_AUTH_PASS=
+DATABASE_URL=horizon_test
```
jonallured, line 2: I remember while we were pairing we talked about the value of `DATABASE_URL` and its relationship to one's local databases but I might not have done a great job of explaining myself! 😛 

By default a Rails application will use the database configuration found in the `config/database.yml` file. Only when a different type of setup is required should another configuration method be used. Setting `DATABASE_URL` is how we configure things in k8s, but I would NOT expect to see this in a local setup.

The docs do a pretty decent job running through things:

https://guides.rubyonrails.org/configuring.html#configuring-a-database

> If you have both config/database.yml and ENV['DATABASE_URL'] set then Rails will merge the configuration together.

Looking back at the config/database.yml file on master, I would expect it to work great for local development, including running tests! The wrinkle here is that if one has DATABASE_URL set in their .env file, then the test one will pick that up!! This is super dangerous!

And so what you've done here is great! Including the `DATABASE_URL` in the test version of this file prevents this particular mistake from occurring, but I just want to emphasize that using this environment var at all is super risky. A better approach to using staging/production data is the pull data task over in the bin folder. Pulling data from a remote DB into a local one ensures that while messing around with things, nothing upstream is affected.

Anyway, thanks for coming to my TED talk!! ❤️ ❤️ ❤️

app/controllers/projects_controller.rb
```diff
@@ -1,5 +1,14 @@
...
     @presenter = ProjectsPresenter.new(params)
+    @props = {
+      tags: tags
+    }
+  end
+
+  private
+
+  def tags
+    Project.all.pluck(:tags).flatten.compact.uniq.sort
```
jonallured, line 12: As we discussed, it's possible there's a better way to model this - I'm envisioning a `Team` model rather than a tag string column. But yeah totally optional and for a future PR.

app/controllers/projects_controller.rb
```diff
@@ -1,5 +1,14 @@
 class ProjectsController < ApplicationController
   def index
     @presenter = ProjectsPresenter.new(params)
+    @props = {
+      tags: tags
```
jonallured, line 5: I'm surprised to not see a `projects` key here. 🤔 Is that `@presenter` ivar used somewhere?

app/helpers/projects_helper.rb
```diff
@@ -39,7 +39,8 @@ def gravatar_from_log_line(line)
   end
 
   def first_name_from_log(line)
-    ReleasecopService.parsed_log_line(line)[:name].split(' ')[0]
+    name = ReleasecopService.parsed_log_line(line)[:name]
+    name && name.split(' ')[0] || ""
```
jonallured, line 43: Nice refactor. 💪

app/javascript/components/__tests__/MainLayout.test.tsx
```diff
@@ -1,11 +1,20 @@
...
 
 describe("MainLayout", () => {
   it("Renders the nav bar", () => {
     const component = mount(<MainLayout />)
     expect(component.find(NavBar)).toHaveLength(1)
     expect(component.text()).toMatch("Horizon")
+    expect(component.text()).not.toMatch("Teams")
+  })
+
+  it("Renders 'teams' dropdown if tags are provided", () => {
```
jonallured, line 14: Nice specs

app/views/projects/index.html.erb
```diff
@@ -1,2 +1,10 @@
+<%= content_tag :div,
+  id: "projects_data",
+  data: {
+    props: @props.to_json
+  } do %>
+<% end %>
+
+<%= javascript_pack_tag 'projects' %>
```
jonallured, line 8: Ohhh, I get it now - this PR does not replace the entire screen with React, just the top part. You'll come back later to add more stuff to props and take over the rest of the screen, right?

I think there's a slightly easier way to write this DIV though, maybe something like this:

```erb
<div id="projects_data" data-props="<%= @props.to_json %>"></div>
```

Or something like that, my ERB skills have atrophied since discovering HAML. 😝

.env.test
eessex: I am guilty of connecting to staging locally, and want to guard against anyone who sets that locally -- going to add a comment warning not to use it.

app/controllers/projects_controller.rb
eessex: Not going to change models just yet, but agreed, `tags` is misleading for how we use it.

app/controllers/projects_controller.rb
eessex: Will send a few more PRs over the coming week to add project views!

jonallured: Woot!

.env.test
jonallured: Perfect!

jonallured:
Applied MOG label, let's do this.

PR #118, opened 2020-06-17
Add project show view

eessex:
Adds a show view (`/projects/id`) to consolidate the information we have about a project, and provide links to edit/deploy blocks.

Screen Shot 2020-06-17 at 1 05 16 PM


app/javascript/apps/Project/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+                </Sans>
+                <EditIcon />
+              </Flex>
+            </Link>
+          </Box>
+
+          {gitRemote && (
+            <Box my={1}>
+              <Sans size="3">
+                <Link href={gitRemote}>{gitRemote}</Link>
```
joeyAghion, line 66: This looks like it will include the `.git` extension which isn't actually navigable. You could use a helper to transform it though.

joeyAghion: My mistake! It looks like Github redirects these paths successfully.

eessex:
Just added JS tests to this and marked as ready for review. I haven't added rspec tests for the `ProjectPresenter`, purely because I'm a bit unfamiliar with rspec patterns. I'm going to add a follow up for this as a separate PR - logic is pre-existing and moved out of `ProjectsPresenter`, where it is currently in production untested.

eessex:
@jonallured thanks for your thoughts on testing -- I'll put a quick pair session on your cal to learn more about rspec suggestions. Thanks!

PR #120, opened 2020-06-18
Use react for frontend, projects index

eessex:
Converts projects index to react/palette, our front-end is now fully react.
Project summaries now link through to the new project view (#118)

Keeping our css pipeline around just for the "aged" look, but deleted all other rails front-end files

Screen Shot 2020-06-18 at 3 41 58 PM


app/javascript/components/Project/__tests__/ProjectSummary.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+      ...unreleasedProjectFixture,
+    }
+  })
+  it("Renders project name and description", () => {
+    const component = getWrapper()
+    expect(component.text()).toMatch("Force")
+    expect(component.text()).toMatch("Artsy.net web front-end")
+  })
+
+  it("Renders 'bocked' if blocks", () => {
```
anandaroop, line 23:
```suggestion
  it("Renders 'blocked' if blocks", () => {
```

app/presenters/projects_presenter.rb
```diff
@@ -17,46 +18,14 @@ def released_projects
...
-  end
-
   private
 
   def projects
     @projects ||= begin
       query = Project.includes(:stages, snapshot: [:comparisons]).where(@params.permit(:organization_id))
       query = query.where('tags ?| array[:tags]', tags: @params[:tags]) if @params[:tags]&.any?
-      query.map { |p| ProjectWrapper.new(p) }
+      query.map { |p| ProjectPresenter.new(p) }
```
anandaroop, line 27: Minor readability nit — maybe `query.entries.map` or `query.to_a.map` ?

spec/features/projects_spec.rb
```diff
@@ -19,20 +19,23 @@
...
-
-    expect(page).to have_content('Out of sync')
-    expect(page).to have_content('Foo Php')
-    expect(page).to have_content('Shipping')
-    expect(page).to have_content('Scheduling')
+    expect(page).to have_selector 'div[id="projects_data"]'
+    props = find('div[id="projects_data"]')["data-props"].as_json
+    expect(props).to have_content('Foo Php')
+    expect(props).to have_content('Shipping')
+    expect(props).to have_content('Scheduling')
```
anandaroop, line 26: Is it no longer possible to assert against the rendered page content given these changes?

app/presenters/projects_presenter.rb
```diff
@@ -1,8 +1,9 @@
 class ProjectsPresenter
-  VIEWS = %w[detail dashboard]
-
   def initialize(params)
     @params = params
+    @released_projects = released_projects()
+    @unreleased_projects = unreleased_projects()
+    @detailed_projects = detailed_projects()
```
joeyAghion, line 6: Rubocop (whom I intentionally never invited to this repo) would complain about the unnecessary `()` in these lines. More generally, I'm not sure why these 3 instance variables are being set at all. I don't see references to them, so they might just be duplicating work.

spec/features/projects_spec.rb
eessex: I'm guessing this relates to hydration, I don't think we are server side rendering. All the view logic is tested on the Jest side, so repurposed this test to confirm that `props` are present and that actioncable updates are pushed to props.

app/presenters/projects_presenter.rb
eessex: These are used in the controller here: https://github.com/artsy/horizon/pull/120/files/21ea2acc6c3d123e0ff4c6ed907072621b322404#diff-757cba42ec9599bac53d4cc33cdde3bfR6

I was planning to add rubocop, my Ruby syntax is pretty rough so it seems like it would help me out -- I wonder if there was a specific reason you excluded it?

app/javascript/apps/Projects/index.tsx
```diff
@@ -1,11 +1,64 @@
...
+  releasedProjects,
+  unreleasedProjects,
+  tags,
+  params,
+}) => {
+  return (
+    <MainLayout tags={tags}>
+      <Box px={3} pt={2}>
+        {params.tags && (
+          <Sans size="8" pb={4} style={{ textTransform: "capitalize" }}>
```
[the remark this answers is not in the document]
eessex: Maybe a matter of preference, but it seems to work with all the tags that exist so far -- I find it a bit nicer visually:
Screen Shot 2020-06-24 at 1 40 08 PM

app/presenters/projects_presenter.rb
joeyAghion: I believe the controller is calling the methods directly, not using the instance variables.

Re. rubocop, it's probably fine to add now that more people are contributing to this repo. I agree with 95% of what it does but the 5% occasionally grates on me.

eessex: Yes you're right! Will remove this.

spec/features/projects_spec.rb
anandaroop: Yeah, _if_ we wanted to get this working we'd likely need to set up JS testing in Capybara using whatever helpers are current. (A year ago it was the [Webdrivers](https://github.com/titusfortner/webdrivers) gem, not sure currently).

jonallured: Yeah, would love to work on improving this js browser testing at some point! Once the first one is setup, it's really not too bad to add additional full outer-loop tests and they can be quite helpful.

(And yes, that webdrivers gem is still the best practice these days @anandaroop )

PR #121, opened 2020-06-18
Add project data service

eessex:
Proof of concept to fetch important repo meta/info from github. 

The objective is to save data fetched from github to projects at a weekly cadence, to keep our basic meta up to date. These fields are not exposed in active admin. 

This is a step towards retiring the notion project list in favor of an auto-updated version in horizon.

- Adds `Dependencies` table, a `Project` has many `Dependencies`.
- Adds fields `renovate`, `ci_provider`, and `orbs` to `Project`
- Adds `ProjectDataService` to fetch and update data on `Projects`
- Adds rake task and cron task to update project data

Returns the following for... 
Force: `:ci_provider=>"circleci", :renovate=>true, :orbs=>["hokusai", "yarn"]}`
Gravity: `:ci_provider=>"circleci", :renovate=>true, :orbs=>["hokusai"]}`

And creates new dependencies on `Projects` for Ruby and Node:
Screen Shot 2020-06-22 at 4 43 31 PM

Screen Shot 2020-06-24 at 5 16 35 PM


app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def initialize(project, access_token)
+    @project = project
+    @access_token = access_token
+    @client = Octokit::Client.new(access_token: access_token)
+    @circle_config = circle_config
+  end
+
+  def github_org
+    Organization.find(@project.organization_id).name
```
joeyAghion, line 13: This can just be `@project.organization.name`.

Related: this isn't the only place we assume an `Organization` name matches a Github org name. We also do it here: https://github.com/artsy/horizon/blob/88dabc801c01d1bc58ebe4b8d1b9f9911063d444/app/models/deploy_strategy.rb#L21

It might be good to encapsulate that somewhere like `project.github_org_name` (and use the new method in both places) so we have the option to improve it in the future (like by basing it on the actual git remote).

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def project_name
+    url = git_remote.split("/").last
+    url.gsub(".git", "")
+  end
```
joeyAghion, line 19: Ditto about this... the line linked above is performing this same operation and they can probably be combined.

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def git_remote
+    stage = @project.stages&.detect{ |s| s.name == "master" }
+    stage && stage.git_remote
```
joeyAghion, line 23:
```suggestion
    stage&.git_remote
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def computed_properties
+    {
+      base_libraries: {
+        ruby: ruby_version,
+        node: node_version
+      },
```
joeyAghion, line 31: These are more "runtimes" than libraries. Rails and express and so on would be libraries though.

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+      ci_provider: ci_provider,
+      renovate: has_renovate?,
+      orbs: orbs
+    }
+  end
+
+  def ci_provider
+    @circle_config && "circleci"
```
joeyAghion, line 39:
```suggestion
    'circleci' if @circle_config
```

(Equivalent but slightly more ruby-ish.)

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def has_renovate?
+    renovate_config && true
```
joeyAghion, line 43:
```suggestion
    !!renovate_config
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def orbs
+    orbs = []
+    @circle_config.include?("artsy/hokusai@") && orbs.push("hokusai")
+    @circle_config.include?("artsy/yarn@") && orbs.push("yarn")
```
joeyAghion, line 49:
```suggestion
    orbs.push("hokusai") if @circle_config&.include?("artsy/hokusai@")
    orbs.push("yarn") if @circle_config&.include?("artsy/yarn@")
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+    orbs
+  end
+
+  def ruby_version
+    file = fetch_github_file('.ruby-version')
+    return if !file
+    decode_content(file) || "unknown verion"
```
joeyAghion, line 56:
```suggestion
    decode_content(file) || 'unknown version'
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def node_version
+    file = fetch_github_file('package.json')
+    return if !file
+    json = JSON.parse(decode_content(file))
+    json["engines"]["node"] || "unknown verion"
```
joeyAghion, line 63:
```suggestion
    json['engines']['node'] || 'unknown version'
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def circle_config
+    file = fetch_github_file('.circleci/config.yml')
+    return if !file
+    json = decode_content(file)
+    json
```
joeyAghion, line 70:
```suggestion
    decode_content(file)
```

In this case the file isn't actually "json." Either way, I don't think the variable is necessary.

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def renovate_config
+    file = fetch_github_file('renovate.json')
+    return if !file
+    json = decode_content(file)
+    json
```
joeyAghion, line 77:
```suggestion
    decode_content(file)
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def decode_content(file)
+    Base64.decode64(file[:content]).gsub(/\n/, '')
+  end
+
+  def fetch_github_file(path)
+    file = @client.contents("#{github_org}/#{project_name}", :path => path)
+    file && file.to_h
+    rescue Octokit::NotFound
```
joeyAghion, line 87:
```suggestion
  rescue Octokit::NotFound
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def renovate_config
+    file = fetch_github_file('renovate.json')
+    return if !file
+    json = decode_content(file)
+    json
+  end
+
+  def decode_content(file)
+    Base64.decode64(file[:content]).gsub(/\n/, '')
```
joeyAghion, line 81: Why the `gsub`? I was curious about this and tried a few files and they seemed to decode fine without this.

app/services/project_data_service.rb
joeyAghion: If you wanted to keep this general (runtimes, libraries, etc.) you might just call them "dependencies."

app/services/project_data_service.rb
eessex: I noticed this specifically on `.ruby_version`, where line breaks were consistently included. Because I am saving this data as strings I wanted to be cautious not to inadvertently save breaks.

eessex:
@joeyAghion this has been updated and is ready for another review.  Comparing a dependency version against an expectation for the org will be covered in a separate PR.

app/services/project_data_service.rb
```diff
@@ -0,0 +1,111 @@
...
+    node_version && update_dependency('node', node_version)
+  end
+
+  def update_dependency(name, version)
+    d = @project.dependencies.where(name: name)
+    if d.empty?
+      @project.dependencies << Dependency.create(name: name, version: version)
+    else
+      d.update(name: name, version: version)
+    end
```
joeyAghion, line 59: Activerecord offers `find_or_initialize_by` which could simplify this method to:

```ruby
@project.dependencies.find_or_initialize_by(name: name).update!(version: version)
```

app/services/project_data_service.rb
```diff
@@ -0,0 +1,111 @@
...
+  def circle_config
+    file = fetch_github_file('.circleci/config.yml')
+    return unless file
+
+    decode_content(file)
+  end
+
+  def travis?
+    file = fetch_github_file('.travis.yml')
+    true if file
```
joeyAghion, line 91:
```suggestion
    !!fetch_github_file('.travis.yml')
```

hokusai/production.yml
```diff
@@ -164,7 +164,7 @@ spec:
 apiVersion: batch/v2alpha1
 kind: CronJob
 metadata:
-  name: horizon-refresh-cron
+  name: horizon-refresh-comparisons-cron
```
joeyAghion, line 167: Note that I think renaming the cron will create a _new_ cron upon the next deploy, and the old one will need to be manually deleted. Ditto in staging.

hokusai/staging.yml
```diff
@@ -190,3 +190,34 @@ spec:
...
+  name: horizon-refresh-project-data-cron
+spec:
+  schedule: "0 0 * * 1"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: horizon-refresh-project-data-cron
```
joeyAghion, line 206: "Data" is a little general-sounding. I might call this task and the cron `refresh_dependencies` or `refresh_components` if you want it to imply more than dependencies (like CI).

app/presenters/project_presenter.rb
```diff
@@ -94,9 +94,10 @@ def gravatar_from_log_line(line)
   def as_json(_options = nil)
     attributes = @project.as_json
     computed_attributes = {
+      block: block,
       comparedStages: compared_stages,
+      dependencies: @project.dependencies,
```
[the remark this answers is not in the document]
eessex: They are serialized already, I think because they are wrapped in `as_json`.

hokusai/production.yml
eessex: Thanks for heads up -- wanted to make these names more specific now that we have multiple, will add #migration steps.

PR #129, opened 2020-06-25
Add Rubocop

eessex:
Adds rubocop to project, default rake task, and runs on all files.


app/presenters/project_presenter.rb
```diff
@@ -22,12 +25,12 @@ def severity
...
-    stage && stage.git_remote
+    stage = ordered_stages&.detect { |s| s.name == 'master' }
+    stage&.git_remote
   end
 
-  def is_kubernetes?
-    hokusai_stages = ordered_stages&.detect{ |s| s.hokusai }
+  def kubernetes?
+    hokusai_stages = ordered_stages&.detect { |s| s.hokusai }
     !hokusai_stages.nil?
```
joeyAghion, line 34: This method might be simplified to `ordered_stages&.any?(&:hokusai)`.

app/presenters/project_presenter.rb
```diff
@@ -37,9 +40,9 @@ def block
   end
 
   # enumerates pairs of stages, the corresponding comparison object, and severity score
-  def compared_stages
+  def compared_stages # rubocop:disable Metrics/CyclomaticComplexity
```
joeyAghion, line 43: Surprised about this too but 🤷

app/presenters/project_presenter.rb
```diff
@@ -98,10 +105,10 @@ def as_json(_options = nil)
       gitRemote: git_remote,
       block: block,
       isFullyReleased: fully_released?,
-      isKubernetes: is_kubernetes?,
+      isKubernetes: kubernetes?,
```
joeyAghion, line 108: If possible, I'd prefer the client-side names to match the server-side ones (aside from casing).

app/services/comparison_service.rb
```diff
@@ -44,7 +46,7 @@ def self.refresh_comparisons_for_organization(org)
     new_snapshots
   end
 
-  def refresh_comparisons
+  def refresh_comparisons # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
```
joeyAghion, line 49: This one is understandable 😅

app/services/releasecop_service.rb
```diff
@@ -1,7 +1,9 @@
+# frozen_string_literal: true
+
 class ReleasecopService
   attr_accessor :project
 
-  LOG_LINE_EXPR = /^(?<sha>[0-9a-f]+) (?<date>[0-9\-]+) (?<message>.*) \((?<name>.*), (?<email>.*)\)\w*$/ # %h %ad %s (%an, %ae)
+  LOG_LINE_EXPR = /^(?<sha>[0-9a-f]+) (?<date>[0-9\-]+) (?<message>.*) \((?<name>.*), (?<email>.*)\)\w*$/.freeze
```
joeyAghion, line 6: We should keep the comment, or maybe just link to the git-log expression that it must match: https://github.com/joeyAghion/releasecop/blob/bc9c108f1c4ff33fefe75ffb0c44f5ac8822594d/lib/releasecop/comparison.rb#L5

app/presenters/project_presenter.rb
eessex: This one is weird -- `rubocop` says i need to use `kubernetes?` instead of `is_kubernetes?`, however it is JS convention to preface booleans with `isFoo`... open to changing but could also disable the rule here.

app/models/deploy_strategy.rb
```diff
@@ -23,13 +25,13 @@ def github_repo
...
 
-  def validate_arguments
+  def validate_arguments # rubocop:disable Metrics/CyclomaticComplexity, Metrics/AbcSize
     return unless PROVIDERS.include?(provider) # don't bother if provider invalid
 
     unless REQUIRED_ARGUMENTS[provider].all? { |a| (arguments || {}).keys.include?(a) }
       errors.add(:arguments, "must include #{REQUIRED_ARGUMENTS[provider].to_sentence}")
     end
-    unless (arguments || {}).keys.all? { |a| SUPPORTED_ARGUMENTS[provider].include?(a) }
+    unless (arguments || {}).keys.all? { |a| SUPPORTED_ARGUMENTS[provider].include?(a) } # rubocop:disable Style/GuardClause
```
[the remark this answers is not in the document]
eessex: added the guard statement but still erroring on `validate_arguments`.

app/presenters/project_presenter.rb
joeyAghion: We should definitely keep `kubernetes?` on the ruby side, where the convention is to use `?` names instead of `is`. Fine with me if you like the `is` convention on the JS side, as long as we're consistent.

app/presenters/project_presenter.rb
yuki24: Personally I'd always disable `Metrics/CyclomaticComplexity`. It leads to unnecessary refactoring and code massaging for the sake of getting a good rate.

app/presenters/project_presenter.rb
```diff
@@ -22,13 +25,12 @@ def severity
...
-    stage && stage.git_remote
+    stage = ordered_stages&.detect { |s| s.name == 'master' }
+    stage&.git_remote
   end
 
-  def is_kubernetes?
-    hokusai_stages = ordered_stages&.detect{ |s| s.hokusai }
-    !hokusai_stages.nil?
+  def is_kubernetes? # rubocop:disable Naming/PredicateName
+    true if ordered_stages&.detect { |s| !s.hokusai&.empty? }
```
yuki24, line 33:
```suggestion
    ordered_stages&.any? { |s| !s.hokusai&.empty? }
```

Not sure if returning `nil` is specifically important, but the `#any` method could make it simpler a bit.

.rubocop.yml
```diff
@@ -0,0 +1,14 @@
+AllCops:
+  NewCops: enable
+  Exclude:
+    - db/**/*
+    - bin/**/*
+    - node_modules/**/*
+    - spec/*.rb
```
yuki24, line 7: I think listing spec files makes sense too. Is there any reason it needs to be excluded?

PR #143, opened 2020-07-01
Dependencies are evaluated against expected version, maintenance messages

eessex:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2183

- Adds support to evaluate if a dependency is up to date by comparing against  an expected version. 
- Set version expectations with env vars `EXPECTED_VERSION_RUBY` and `EXPECTED_VERSION_NODE`
- Display status icon on `dependencies` row in `ProjectList` to visualize if maintenance is required
- Display a list of suggested maintenance messages on `Project` page

Screen Shot 2020-07-01 at 4 47 38 PM
Screen Shot 2020-07-01 at 4 11 46 PM


app/javascript/Typings.ts
```diff
@@ -31,6 +31,7 @@ export interface Dependency {
   name: string
   project_id: number
   version: string
+  update_required?: boolean
```
joeyAghion, line 34: This casing is inconsistent with the JS/JSON-style casing elsewhere in this file. Intentional?

app/javascript/apps/Project/index.tsx
```diff
@@ -15,8 +15,9 @@ import { Project, Stage, Tags as TagsType } from "Typings"
 import { deployBlockPath, projectEditPath } from "../../shared/UrlHelper"
 import { formattedDependencies, formattedTags } from "../../shared/Helpers"
 import { MainLayout } from "../../components/MainLayout"
-import { StageWithComparison } from "../../components/Stage/StageWithComparison"
+import { ProjectMaintenanceRecomendations } from "../../components/Project/MaintenanceRecommendations"
```
joeyAghion, line 18: Typo (should be`ProjectMaintenanceRecommendations`) here and in a few related locations.

app/services/project_data_service.rb
```diff
@@ -52,8 +52,26 @@ def update_dependencies
...
+
+    current = version.delete('^0-9.').split('.')
+    current_major = current[0]&.to_i
+    current_minor = current[1]&.to_i
+    return unless current_major
+
+    return true if current_major < expected_major
+    return true if expected_minor && current_minor < expected_minor
+
+    false
```
joeyAghion, line 74: Comparing semantic version strings is non-trivial and we might be able to take advantage of prior art, such as in `Gem::Version` or [`semantic`](https://github.com/jlindsey/semantic). This would handle special cases like betas, and maybe even support matching against multiple constraints (like minimum _and_ maximum versions).

.env.test
```diff
@@ -3,3 +3,5 @@ BASIC_AUTH_PASS=
 # We shouldn't need this, is an extra safeguard for
 # developers working against staging.
 DATABASE_URL=horizon_test
+EXPECTED_VERSION_RUBY=2.6
+EXPECTED_VERSION_NODE=12
```
joeyAghion, line 7: Based on how these operate now, I might call these `MINIMUM_...` rather than `EXPECTED_...`.

app/services/project_data_service.rb
```diff
@@ -52,8 +52,26 @@ def update_dependencies
...
 
   def update_dependency(name, version)
+    update_required = dependency_update_required(name, version)
     @project.dependencies.find_or_initialize_by(name: name)
-            .update!(version: version)
+            .update!(version: version, update_required: update_required)
+  end
+
+  def dependency_update_required(name, version)
+    expectation = Horizon.config.stringify_keys["expected_version_#{name}"].split('.')
```
joeyAghion, line 61: This seems like it would fail upon empty config. This is fine if you end up going with my suggestion to implement defaults directly in `_config.rb`, but otherwise should have a guard.

(Minor: by convention this method's name should also end in `?`.)

config/initializers/_config.rb
```diff
@@ -4,5 +4,6 @@
 
 Horizon.config = {
   basic_auth_user: ENV['BASIC_AUTH_USER'] || 'admin',
-  basic_auth_pass: ENV['BASIC_AUTH_PASS']
+  expected_version_ruby: ENV['EXPECTED_VERSION_RUBY'],
+  expected_version_node: ENV['EXPECTED_VERSION_NODE']
```
joeyAghion, line 8: This might be a better place to set defaults than the `.env.test` file, so tests can be run locally as well as by hokusai.

eessex: I don't want to set the version in code so that it can change on the fly, the `.env.test` is used just for comparisons in tests.

joeyAghion: You can still update the configuration, and any tests depending on specific values should be mocking them anyway like:

```ruby
allow(Horizon).to receive(:config).and_return(expected_version_ruby: '...')
```

Any defaults here would just be a fall-back.

app/services/project_data_service.rb
eessex: I was able to get this halfway with the `semantic` gem-- unfortunately we use a variety of conventions in our project's version files, and dependencies frequently fail with errors like this: `12.x.x is not a valid SemVer Version (http://semver.org)`.

I updated for it to use `semantic` first to compare when `current` is a valid SemVer, and left my more manual parsing for the moment. If we enforce some naming conventions we could potentially remove this, but most of our versions fail without.

app/javascript/Typings.ts
eessex: We are not parsing this data after returning to the db, just sending as json so expected. All the values we are creating/manually assigning to send to the client do use the camel convention.

app/javascript/components/Projects/ProjectsList.tsx
```diff
@@ -161,10 +177,16 @@ export const ProjectsListRow: React.FC<{ project: Project }> = ({
...
         {orbs?.length > 0 ? (
           <Flex my={1} alignItems="center">
-            <Tags tags={formattedTags(orbs)} />
+            <Tags tags={formattedOrbs(orbs)} />
           </Flex>
+        ) : isKubernetes ? (
+          <XCircleIcon fill="red100" />
         ) : (
-          isKubernetes && <XCircleIcon fill="red100" />
+          ci_provider !== "circleci" && (
```
jonallured, line 185: I would love to see a helper function extracted here for this conditional logic - nested ternary expressions break my brain! 😝

spec/services/project_data_service_spec.rb
```diff
@@ -71,6 +78,7 @@
       ProjectDataService.new(project).update_dependency('ruby', '2.4.3')
       expect(Dependency.count).to eq(1)
       expect(project.dependencies.first.version).to eq('2.4.3')
+      expect(project.dependencies.first.update_required).to be_truthy
```
jonallured, line 81:
```suggestion
      expect(project.dependencies.first).to be_update_required
```

spec/services/project_data_service_spec.rb
```diff
@@ -79,6 +87,7 @@
       expect(Dependency.count).to eq(2)
       expect(project.dependencies.last.name).to eq('node')
       expect(project.dependencies.last.version).to eq('v12')
+      expect(project.dependencies.first.update_required).to be_falsey
```
jonallured, line 90:
```suggestion
      expect(project.dependencies.first).to_not be_update_required
```

spec/services/project_data_service_spec.rb
```diff
@@ -92,6 +101,33 @@
...
+      needs_update = ProjectDataService.new(project).dependency_update_required?('node', 'v13')
+      expect(needs_update).to be_falsey
+    end
+
+    it 'can handle unknown versions' do
+      needs_update = ProjectDataService.new(project).dependency_update_required?('ruby', 'unknown version')
+      expect(needs_update).to be_falsey
+    end
+  end
+
```
jonallured, line 130:
```suggestion
  describe 'dependency_update_required?' do
    let(:service) { ProjectDataService.new(project) }

    it 'can identify out of date ruby versions' do
      needs_update = service.dependency_update_required?('ruby', '2.4.3')
      expect(needs_update).to be_truthy
    end

    it 'can verify ruby version is up to date' do
      needs_update = service.dependency_update_required?('ruby', '2.6.6')
      expect(needs_update).to be_falsey
    end

    it 'can identify out of date node versions' do
      needs_update = service.dependency_update_required?('node', '>=8.12.x')
      expect(needs_update).to be_truthy
    end

    it 'can verify node version is up to date' do
      needs_update = service.dependency_update_required?('node', 'v13')
      expect(needs_update).to be_falsey
    end

    it 'can handle unknown versions' do
      needs_update = ProjectDataService.new(project).dependency_update_required?('ruby', 'unknown version')
      expect(needs_update).to be_falsey
    end
  end

```

app/services/project_data_service.rb
```diff
@@ -52,8 +54,31 @@ def update_dependencies
   end
 
   def update_dependency(name, version)
+    update_required = dependency_update_required?(name, version)
     @project.dependencies.find_or_initialize_by(name: name)
-            .update!(version: version)
+            .update!(version: version, update_required: update_required)
+  end
+
+  def dependency_update_required?(name, version)
```
jonallured, line 62: This is an interesting method. Unless I'm mistaken, I believe that it does NOT use the state on this class, correct? Like, I'm not seeing a direct relationship between the `project` that this service is instantiated with and the logic of this method. That leads me to believe it might be in the wrong location.

The arguments that are passed in are properties of a `dependency`, correct? If so, then I would expect to see this defined over there:

```rb
test_project = Project.create(name: "test", ...)
ruby_dep = test_project.dependencies.create(name: "ruby", version: "2.7.1", ...)
ruby_dep.update_required?
```

And then that `update_required?` method would take no args because the name and version are on the dependency instance. It could reach into the `Horizon.config` to get required versions and return the boolean.

eessex: Moved this and some additional logic into models, and added `as_json`

PR #152, opened 2020-07-14
Project criticality

eessex:
- Adds column `criticality` to `Projects`
- Adds UI for `criticality` to active admin
- `/projects` can filter on `criticality` query param
- Project list and project show display criticality

Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2103


app/presenters/projects_presenter.rb
```diff
@@ -22,6 +22,7 @@ def released_projects
   def projects
     @projects ||= begin
       query = Project.includes(:stages, snapshot: [:comparisons]).where(@params.permit(:organization_id))
+      query = query.where(criticality: @params[:criticality]) if @params[:criticality]
```
joeyAghion, line 25: You might be able to just add `:criticality` to the `permit` args in the line above and leave out this line.

db/migrate/20200713215815_add_criticality_to_projects.rb
```diff
@@ -0,0 +1,5 @@
+class AddCriticalityToProjects < ActiveRecord::Migration[6.0]
+  def change
+    add_column :projects, :criticality, :integer, :default => 1
```
joeyAghion, line 3: I'd leave the `default` out, unless you actually want to associate everything in the database with `1`. It might be easier and more visible to leave them `NULL`/unset until they're updated.

app/javascript/components/Projects/ProjectsList.tsx
```diff
@@ -217,3 +226,32 @@ const AutoDeployIcon: React.FC<Project> = (props) => {
...
+    case 0:
+      name = "0: Unsupported"
+      break
+  }
+
+  return (
+    <Tags
+      tags={[
+        {
+          href: `/?criticality=${criticality}`,
```
joeyAghion, line 251: Shouldn't there be a `projects` in this URL?

app/models/project.rb
```diff
@@ -60,6 +60,7 @@ def as_json(_options = nil)
     {
       block: block,
       ciProvider: ci_provider&.titleize,
+      criticality: criticality,
```
joeyAghion, line 63: Since the forms only support certain values for the new `criticality` field, it would make sense to also add a model validation enforcing that to this class. Something like:

```ruby
validates :criticality, inclusion: { in: [0, 1, 2, 3] }
```

app/javascript/components/Projects/ProjectsList.tsx
eessex: Root automatically resolves to projects so essentially the same.

PR #159, opened 2020-07-17
Development fixes and small release-assignment improvement

joeyAghion:
Please review by commit, since I had to modify a few things to get this work started. These steps included:
* Some developer environment fixes to be able to run locally.
* Refactored `DeployService` operations into instance methods for better encapsulation and memoization.
* If a release PR already exists, but has no assignee, check if there's a good candidate and assign them.

This^ last problem was my reason for coming here. Often, the initial diff in a release PR is contributed by bots, and human contributors' commits get added later. Until now, Horizon only chose an assignee when creating the release PR (so based only on those initial commits). Going forward, it will add an assignee to an existing release PR when human contributions get added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,16 @@ Visual representations of release pipelines.
 
 ## Setup
 
-    # via hokusai
+With docker:
+
     hokusai dev run 'bundle exec rake db:migrate'
     hokusai dev start
-    # or
-    dotenv bundle exec rails db:prepare
-    dotenv bundle exec rails server
+
+Or on localhost:
+
+    bundle exec rails db:prepare
+    yarn install
+    bundle exec rails server
 
     # run the webpack-dev-server in a seperate terminal window for hot reloading:
     ./bin/webpack-dev-server
--- a/spec/features/comparisons_spec.rb
+++ b/spec/features/comparisons_spec.rb
@@ -57,7 +57,7 @@ RSpec.feature 'Comparisons', type: :feature do
       allow_any_instance_of(Releasecop::Checker).to receive(:check).and_return(
         Releasecop::Result.new('shipping', [large_comparison])
       )
-      expect(DeployService).to receive(:start)
+      expect_any_instance_of(DeployService).to receive(:start)
       ComparisonService.new(project).refresh_comparisons
     end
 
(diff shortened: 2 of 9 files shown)


spec/features/comparisons_spec.rb
```diff
@@ -57,7 +57,7 @@
       allow_any_instance_of(Releasecop::Checker).to receive(:check).and_return(
         Releasecop::Result.new('shipping', [large_comparison])
       )
-      expect(DeployService).to receive(:start)
+      expect_any_instance_of(DeployService).to receive(:start)
```
starsirius, line 60: 🙈

spec/features/deploy_spec.rb
```diff
@@ -68,4 +68,25 @@
...
   end
+
+  it 'adds assignees to existing deploy PRs when unassigned' do
+    expect_any_instance_of(Octokit::Client).to receive(:create_pull_request)
+      .with('artsy/candela', 'release', 'staging', anything, anything)
+      .and_raise(Octokit::UnprocessableEntity)
+    allow_any_instance_of(Octokit::Client).to receive(:pull_requests)
+      .with('artsy/candela', base: 'release', head: 'staging')
+      .and_return([double(number: 42, assignee: nil)])
+    allow_any_instance_of(Octokit::Client).to receive(:pull_request_commits)
```
starsirius, line 79: #minor. Maybe we can always use `expect_...` to be stricter?

README.md
```diff
@@ -18,12 +18,16 @@ Visual representations of release pipelines.
...
     hokusai dev start
-    # or
-    dotenv bundle exec rails db:prepare
-    dotenv bundle exec rails server
+
+Or on localhost:
+
+    bundle exec rails db:prepare
+    yarn install
+    bundle exec rails server
```
starsirius, line 30: Looks like local development doesn't need any special/non-default env variables?

joeyAghion: Correct that's the intention.

spec/features/deploy_spec.rb
joeyAghion: I didn't want to concern this spec too much with _how_ the assignee is chosen, only that there is one.

PR #162, opened 2020-07-19
Remove unnecessary config and update readme

starsirius:
I think my [suggestion](https://github.com/rails/webpacker/issues/1190#issuecomment-358905190) in c62d7e56978063d264670238e9a4e5190f5855b5 to fix the error below was incorrect.
```
ActionController::RoutingError (No route matches [GET] "/packs/projects-8a758bbeb007e45b4351.js")
```
`config.public_file_server.enabled` is already [by default `true`](https://github.com/rails/rails/blob/v6.0.3/railties/lib/rails/application/configuration.rb#L36) in development. And in fact, the config is wrapped in a condition for [dev caching](https://github.com/artsy/horizon/blob/162420dd4c968e2a83c639b21a112f4b456a0927/config/environments/development.rb#L17-L31) which is also by default disabled. Therefore, I don't think it's making a difference. (The linked issue was for production environment, in which `config.public_file_server.enabled` is set to `false` by default.)

Back to the original problem, I actually can't reproduce it on my local. Either having `webpack-dev-server` running or not, I'm able to compile the assets and load the page. Curious if you can still reproduce it. 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Visual representations of release pipelines.
 
 - [Releases: dashboard view](https://releases.artsy.net/projects?organization_id=1&view=dashboard)
 - [Releases: detailed view](https://releases.artsy.net/projects?organization_id=1)
-- [View deploy blocks](https://releases.artsy.net/admin/deploy_blocks) or [create a new one](https://releases.artsy.net/admin/deploy_blocks)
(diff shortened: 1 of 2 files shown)


joeyAghion:
Just tried the following after removing that line from `development.rb`:

```
rm -rf node_modules/ public/packs/
yarn install
bundle exec rails s
```

...and it worked to serve assets locally. I'm not sure what the previous issue was. (I experimented with skipping `yarn install` but the error message is different.) 🤷‍♂️

PR #160, opened 2020-07-19
Add webpack bundle analyzer as dev dependency

starsirius:
Similar to https://github.com/artsy/volt/pull/3512, this adds [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) to dev dependencies for visualizing webpack bundles and identifying optimization opportunities. From the result, there are definitely room for improvement.

I also adds 2 scripts for easier generation of the visualization.

- To generate visualization for development environment, run
  ```
  yarn bundle:stats && yarn bundle:analysis
  ```
  Screen Shot 2020-07-18 at 10 10 00 PM

- To generate visualization for production environment, run
  ```
  RAILS_ENV=production NODE_ENV=production yarn bundle:stats && yarn bundle:analysis
  ```
  Screen Shot 2020-07-18 at 10 11 53 PM

PR #163, opened 2020-07-19
Remove jQuery

starsirius:
Following https://github.com/artsy/horizon/pull/160, another low-hanging fruit is to remove jQuery, which seems to be used in only 1 place. This reduces the JS bundle size by about 300 KB in development (30KB in production and gzipped).

## Before

See jQuery in the `application-*.js`.

Screen Shot 2020-07-19 at 10 09 09 AM

## After

Screen Shot 2020-07-19 at 10 07 51 AM

### DIFF
--- a/app/javascript/channels/project_channel.ts
+++ b/app/javascript/channels/project_channel.ts
@@ -1,12 +1,14 @@
 import consumer from "./consumer"
 
 document.addEventListener("DOMContentLoaded", () => {
+  const organization_id_input = document.querySelector(
+    "#organization_subscription_identifier",
+  ) as HTMLInputElement
+
   consumer.subscriptions.create(
     {
       channel: "ProjectChannel",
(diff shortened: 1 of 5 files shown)


app/javascript/channels/project_channel.ts
```diff
@@ -1,23 +1,28 @@
...
-    channel: "ProjectChannel",
-    organization_id: $("#organization_subscription_identifier").val(),
-  },
-  {
-    connected() {
-      // Called when the subscription is ready for use on the server
-    },
-    disconnected() {
-      // Called when the subscription has been terminated by the server
+document.addEventListener("DOMContentLoaded", () => {
```
starsirius, line 3: I think we need to wait for the DOM to be ready before subscribing; otherwise, we won't get the organization id rendered in the DOM.

PR #165, opened 2020-07-20
Add db/seeds.rb for runnable (and re-runnable) seed data

joeyAghion:
This is a small experiment with database-seeding. I haven't personally worked on any project that used seeds extensively, but they're meant for _exactly_ the kinds of situations we've been discussing: easing initial development and creating a known baseline state.

The `db/seeds.rb` file is the standard rails location for seed data. We could break the operations up among separate files, but I thought this was a manageable size to start. What's nice is that this is just ruby code and it can access our application logic, validations, etc. as necessary, so hopefully this doesn't get too boilerplate-y or repetitive.

The net result is that developers can run the `db:seed` rake task locally _or with `hokusai dev ...`_ to get some minimal data into their environment. If they've made changes and wish to reset, the `db:seed:replant` tasks accomplishes that. The faked comparison data ends up looking like:

**TO DO (separately)**
This seed data contains placeholders for sensitive data. A next step will be to load those secrets from a shared location (or just prompt the developer for them) so that the operations depending on them (like git, kubernetes, or AWS-authorized calls) can succeed locally.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,85 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+artsy_org = Organization.create!(name: 'artsy')
+
+github_aws_profile = artsy_org.profiles.create!(
+  name: 'github/aws',
+  basic_username: 'github',
+  basic_password: '<github_token>',
+  environment: {'AWS_ACCESS_KEY_ID' => '<aws_id>', 'AWS_SECRET_ACCESS_KEY' => '<aws_secret>'}
+)
+heroku_profile = artsy_org.profiles.create!(
+  name: 'heroku',
+  basic_username: 'heroku',
+  basic_password: '<heroku_token>'
+)
+
+candela_project = artsy_org.projects.create!(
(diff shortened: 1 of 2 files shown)


joeyAghion:
I've made some progress on the follow-up step of incorporating shared secrets here: https://github.com/joeyAghion/horizon/compare/dev...joeyAghion:secrets

It respects a `~/.artsy/secrets.yml` file if one exists, but uses fallbacks otherwise. Seems to work in both local and docker-based development. (Will PR once this is out of the way.)

CC @icirellik in case I'm abusing docker somehow.

db/seeds.rb
```diff
@@ -5,3 +5,85 @@
...
+  name: 'staging',
+  git_remote: 'https://git.heroku.com/charge-staging.git',
+  profile: heroku_profile
+)
+charge_production_stage = charge_project.stages.create!(
+  name: 'production',
+  git_remote: 'https://git.heroku.com/charge-production.git',
+  profile: heroku_profile
+)
+charge_project.deploy_blocks.create!(description: 'an example deploy block')
```
eessex, line 89: Minor, but it may make sense to put the deploy block on the service with a diff to be closer to a real world example.

joeyAghion: Fair! Updating...

PR #175, opened 2020-07-27
Bad version of sharing config across local development environments

joeyAghion:
This is just a thought checkpoint. I plan to close this and iterate some more.

### Problem

Productive local development depends on multiple running services, each with configuration (keys, URLs...), data (users, partners, access control lists...), networking (hostnames, ports...), and dependencies (databases, caches, language versions...) set up to operate well together. In practice this is difficult to get right, so end-to-end development and QA just doesn't happen or has to wait for multiple changes to reach staging.

### Solution

The aim of [PLATFORM-2640](https://artsyproduct.atlassian.net/browse/PLATFORM-2640) is to make this kind of coordinated set-up easy and automatic. This PR concerns itself with the need for some sensitive or shared values in a service's seed data or ENV. The criteria I have in mind for evaluating solutions include:

* As simple as possible (so in the case of rails backends, reusing built-in `seed` support as in https://github.com/artsy/horizon/pull/165).
* Work when developing on localhost _or_ docker.
* Keep secrets out of repos and docker images.
* Work for open- and closed-source projects.
* Minimize set-up required in each repo (ideally, it's completely optional so doesn't unnecessarily complicate localized work).
* Shouldn't conflict with developers' other activities or across repos (e.g., by clobbering AWS credentials in their env or requiring a certain default language version).
* Since repos use different labels for the same values, and sometimes require different values for similarly-labeled configs, any solution must allow mapping from shared config names to the required in-repo names. E.g., what Volt calls `gravity_client_id` might be `volt_client_id` to Gravity, but should ideally only be present once in the shared data.
* It should be simple to refresh local set-up from an updated source (e.g., as new keys or data get added).

### This PR...

...updates the seed task to respect an optional `~/.artsy/shared.yml` file that provides shared values. In Horizon's simple case, these include Heroku, Github, and AWS keys. Docker-based development accesses these by defining a "volume" that maps the local `~/.artsy` directory to `/home/deploy/.artsy` in the container.

Developers with a `~/.artsy/shared.yml` file get a decent set-up just by running `rake db:seed` or `db:seed:replant`, both locally and with `hokusai dev ...`. My `shared.yml` file looks like:

```yaml
---
aws_access_key_id: REDACTED
aws_secret_access_key: REDACTED
github_token: REDACTED
heroku_token: REDACTED
```

We discussed some reasons why this approach would be problematic:
* Docker-compose volumes come with their own problems, including requiring particular clean-up or not refreshing when expected.
* The `~/.artsy` directory is pretty novel and atypical compared to how we supply most other configuration via ENV.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ The administrative UI can then be found at http://localhost:3000/admin.
 
 To load representative data for development, you have several options:
 * Create organizations, projects, profiles, and stages from the administrative UI.
+* Run the `db:seed` rake task (or `db:seed:replant` to re-seed). The same command can be run via `hokusai dev run ...` for docker-based development.
 * Use the `./bin/pull_data` script to copy data from staging (requires VPN and Artsy developer credentials).
-* Run the `db:seed` rake task (or `db:seed:replant` to re-seed).
 
 Once the cron has run, its snapshots are visible from the `/projects` page.
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -7,7 +7,12 @@
 #   Character.create(name: 'Luke', movie: movies.first)
 
 secrets_path = "#{ENV['HOME']}/.artsy/secrets.yml"
-secrets = File.exist?(secrets_path) ? YAML.load(File.read(secrets_path)) : {}
+secrets = if File.exist?(secrets_path)
+  $stderr.puts "Loading configuration from #{secrets_path}..."
+  YAML.load(File.read(secrets_path))
+else
+  {}
+end
 
 artsy_org = Organization.create!(name: 'artsy')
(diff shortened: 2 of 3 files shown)


joeyAghion:
Another consideration:
* It would be ideal if the same local set-up worked against staging as worked against local backend services. This is just more convenient, since developers are likely to choose the simpler thing of pointing at staging (MP, Gravity, etc.) when changes to those aren't in the scope of their work. However, since staging is on the public internet and has copies of production data, this means that authentication and authorization can't use dummy/guessable values.

joeyAghion:
Closing in favor of https://github.com/artsy/horizon/pull/177

PR #177, opened 2020-07-28
Source shared config from S3 to aid both local and docker development

joeyAghion:
The approach in https://github.com/artsy/horizon/pull/175 depended on developers having a `~/.artsy/shared.yml` with common values. This PR demonstrates how we could instead rely on values from the developer's environment.

Given a developer has a `~/.artsy` file containing the following:

```bash
export ARTSY_ENV=1

export ARTSY_AWS_ACCESS_KEY_ID=REDACTED
export ARTSY_AWS_SECRET_ACCESS_KEY=REDACTED
export ARTSY_GITHUB_TOKEN=REDACTED
export ARTSY_HEROKU_TOKEN=REDACTED
```

...and has a `source ~/.artsy` line somewhere in their `.bash_profile` or `.zshrc` (or whatever), these `ARTSY_`-namespaced values will be available in their environment.

The `bin/setup` script in this repo takes care of initializing a `.env.shared` file with whatever subset the project relies on, possibly with different keys. It also initializes a `.env` file for any local configuration or overrides. (This file is empty in Horizon's case, but a place for overrides is important nonetheless so `.env.shared` can always be cleanly updated later.)

At this point, local development can respect these values using `foreman` like:

```bash
foreman run --env .env.shared,.env bundle exec rails server
```

And docker-based development respects them via the `env_file` declaration.

### What happens when shared or local config changes

When shared configuration changes, the `~/.artsy` file must be updated. This isn't handled in this PR, but that could happen via 1Password, `awscli`, or the team-wide [setup script](https://github.com/artsy/potential/blob/master/scripts/setup) in Potential.

When the project's use of that shared configuration changes, re-running `bin/setup` refreshes the local `.env.shared` file. If necessary, running the built-in `db:seed:replant` task will replace seed data.

### Rejected alternatives

* We could simply list all of the necessary ENV variables in `development.yml`, but this wouldn't be reusable for local development via foreman. In Horizon's case, where the values are only necessary for the seed task and not running the application, it would also be a lot of unfortunate noise in the file. Pushing them into a separate file seemed cleaner.
* Namespacing shared values with `ARTSY_...` is annoying but I really wanted to avoid clobbering other ENV (like `AWS_ACCESS_KEY_ID`).
* It's ugly how the `setup` script writes the `.env.shared` file as a string. It would be nicer and more self-describing for a `.env.shared` file to be committed to the repo and have lines like `FOO=${ARTSY_FOO}`, but I couldn't figure out how to accomplish this without a lot of custom bash-scripting.

Overall I'm worried this pattern is a little fragile. Feedback and ideas for improvement?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,1 @@
-# Place any custom/local configuraton here
+# Place any custom/local configuration here
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+# Place any custom/local configuraton here
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,6 +13,7 @@ README.md
 # Rails
 #
 .env
+.env.shared
 .env.sample
 *.rbc
 capybara-*.html
--- a/.foreman
+++ b/.foreman
@@ -1,0 +1,1 @@
+env: .env.shared,.env
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@
 /config/master.key
 
 .env
+.env.shared
 *.dump
 
 /public/packs
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Visual representations of release pipelines.
 
 ## Setup
 
-The setup script will install prerequisites and create seed data:
+Local development depends on a `.env.shared` file (sourced from S3) and a `.env` file for any overrides or additions. The setup script will initialize these in addition to installing prerequisites and generating seed data:
 
     ./bin/setup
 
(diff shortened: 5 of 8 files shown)


joeyAghion:
Updated based on mob feedback to:
* Specify the `--env .env.shared,.env` options in a `.foreman` file rather than rely on developers to provide them.
* Grab the `.env.shared` directly from S3 instead of depending on developers' environment or writing out the file. This depends on the `aws` command being available and credentials configured, but I think this can safely be handled by the shared setup script in Potential so hopefully isn't a big stretch.

Overall, I find this to be much cleaner and even mergeable.

.env.example
```diff
@@ -0,0 +1 @@
+# Place any custom/local configuration here
```
eessex, line 1: This app doesn't actually use the `.env.example` -- and local config would go in the `.env`, is this file necessary?

eessex: Or rather, it is only used by the setup script-- maybe the script can check for this file to exist rather than including an empty file by default.

joeyAghion: This is a convention I borrowed from other projects, where the `.env.example` serves as a template from which to initialize the `.env` which is `.gitignore`d. While it's true Horizon doesn't need it, I'm hoping this PR can be a reference for other projects that will need it.

PR #183, opened 2020-07-31
Support automatic merging of release PRs after designated period of time

joeyAghion:
One step toward more continuous deployment as described in [PLATFORM-2682](https://artsyproduct.atlassian.net/browse/PLATFORM-2682).

This adds support for an optional `merge_after` argument for `DeployStrategy`s. If a release PR is still open when the specified number of seconds has passed, the PR is merged. This is skipped when unresolved `DeployBlock`s exist thanks to [this existing logic](https://github.com/artsy/horizon/blob/faad9e4b391055ba553103926b53d35c5a83355b/app/services/comparison_service.rb#L59-L62) (and [test](https://github.com/artsy/horizon/blob/faad9e4b391055ba553103926b53d35c5a83355b/spec/features/comparisons_spec.rb#L89)).

This change does _not_ fulfill the requirements described in the ticket around Slack notifications. I figured that could come separately since it will involve new gem dependencies, options, and complexity, plus the automatic merging is completely opt-in for projects in the meantime. I.e., it will only be enabled for the 1 project I test it on, probably Horizon.

### DIFF
--- a/app/services/deploy_service.rb
+++ b/app/services/deploy_service.rb
@@ -27,14 +27,21 @@ class DeployService
 
     assignee = choose_assignee(pull_request)
     github_client.add_assignees(github_repo, pull_request.number, assignee) if assignee
-  rescue Octokit::UnprocessableEntity
-    # PR already exists, so assign if unassigned
+  rescue Octokit::UnprocessableEntity # release PR already exists
     pull_request = github_client.pull_requests(
       github_repo,
       base: deploy_strategy.arguments['base'],
       head: deploy_strategy.arguments['head']
-    ).first
-    if pull_request && pull_request.assignee.blank?
+    ).first || return
+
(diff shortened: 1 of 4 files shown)


app/services/deploy_service.rb
```diff
@@ -27,14 +27,21 @@ def create_github_pull_request
...
       github_repo,
       base: deploy_strategy.arguments['base'],
       head: deploy_strategy.arguments['head']
-    ).first
-    if pull_request && pull_request.assignee.blank?
+    ).first || return
+
+    # merge release PR if merge_after is specified by deploy_strategy
+    if (merge_after = deploy_strategy.arguments['merge_after']) &&
+       Time.now > (pull_request.created_at + merge_after.seconds)
```
eessex, line 39: This might overcomplicate things, but it might be helpful to avoid auto-merging outside of office hours, say nights and weekends.

PR #192, opened 2020-08-05
Optionally notify to Slack prior to merging release PRs

joeyAghion:
Add support for a new `slack_webhook_url` argument for a `DeployStrategy`. If provided, a simple message is sent to the channel specified by the webhook 1 hour before a pending release PR will be automatically merged.

Notes:
* Added `webmock` for making assertions about HTTP requests.
* Because this process runs often (every 10 minutes at the moment), I needed a way to remember if a particular release had already been notified about. I opted to store the URL representing the PR in the `DeployStrategy#arguments` field once it had been notified about. This is a bit of an ugly hack, but avoids embedding any assumptions about cron spacing in case that schedule changes or there are job delays.
* I opted for webhooks (from the many possible forms of Slack authentication) because they grant a specific app access to a specific channel, can be managed individually or app-wide, and are independent of particular user accounts. It also can be accomplished with just a `POST` request so seemed the simplest. To support this I created a "Horizon" app in our Slack workspace and generated an initial webhook (to me). I'll test with that once this is released.

### DIFF
--- a/app/admin/deploy_strategies.rb
+++ b/app/admin/deploy_strategies.rb
@@ -23,7 +23,10 @@ ActiveAdmin.register DeployStrategy do
       f.input :provider, as: :select, collection: DeployStrategy::PROVIDERS
       f.input :profile
       f.input :automatic
-      f.input :arguments_input, label: 'Arguments (JSON)'
+      f.input :arguments_input,
+              label: 'Arguments (JSON)',
+              hint: 'Supported properties: base (branch), head (branch), repo (e.g., org/project), " +
+              "merge_after (sec.), merge_prior_warning (sec., default 3600) slack_webhook_url, warned_pull_request_url'
     end
     f.actions
   end
--- a/app/services/deploy_service.rb
+++ b/app/services/deploy_service.rb
@@ -3,6 +3,8 @@
 class DeployService
   attr_accessor :deploy_strategy
 
+  MERGE_PRIOR_WARNING = 1.hour
+
   def initialize(deploy_strategy)
     self.deploy_strategy = deploy_strategy
   end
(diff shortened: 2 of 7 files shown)


app/services/deploy_service.rb
```diff
@@ -3,6 +3,8 @@
 class DeployService
   attr_accessor :deploy_strategy
 
+  MERGE_PRIOR_WARNING = 1.hour
```
eessex, line 6: Is it possible to set the auto deploy to happen after less than 1 hour?

app/services/deploy_service.rb
```diff
@@ -61,6 +71,20 @@ def github_access_token
       (raise 'A profile and basic_password are required for Github authentication')
   end
 
+  def deliver_slack_webhook(pull_request, webhook_url)
+    uri = URI.parse webhook_url
+    request = Net::HTTP::Post.new(uri.request_uri)
+    request['Content-Type'] = 'application/json'
+    request.body = {
+      text: "The following changes will be released shortly: #{pull_request.html_url}"
```
eessex, line 79: If we are hardcoding the time, we should probably state `in one hour` rather than `shortly`.

app/services/deploy_service.rb
joeyAghion: Technically, yes. Maybe this should just be a default, with an override supported in `arguments`? That would be pretty easy to add.

joeyAghion: (And what should that argument be named? I wasn't happy with this name.)

eessex: not sure, this one is hard - maybe `NOTIFY_BEFORE_DEPLOY_TIME`

app/admin/deploy_strategies.rb
```diff
@@ -23,7 +23,10 @@
       f.input :provider, as: :select, collection: DeployStrategy::PROVIDERS
       f.input :profile
       f.input :automatic
-      f.input :arguments_input, label: 'Arguments (JSON)'
+      f.input :arguments_input,
+              label: 'Arguments (JSON)',
+              hint: 'Supported properties: base (branch), head (branch), repo (e.g., org/project), " +
+              "merge_after (sec.), merge_prior_warning (sec., default 3600) slack_webhook_url, warned_pull_request_url'
```
joeyAghion, line 29: The list of supported arguments was getting long and complex, so I added this inline documentation.

app/services/deploy_service.rb
```diff
@@ -61,6 +76,20 @@ def github_access_token
       (raise 'A profile and basic_password are required for Github authentication')
   end
 
+  def deliver_slack_webhook(pull_request, webhook_url, merge_at)
+    uri = URI.parse webhook_url
+    request = Net::HTTP::Post.new(uri.request_uri)
+    request['Content-Type'] = 'application/json'
+    request.body = {
+      text: "The following changes will be released in #{distance_of_time_in_words_to_now(merge_at)}: #{pull_request.html_url}"
```
joeyAghion, line 84: Used [distance_of_time_in_words_to_now](https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-distance_of_time_in_words_to_now) to specify the amount of time in friendly language, even if custom.

joeyAghion:
Post-release migration will be to update Horizon's [own deploy strategy](https://releases.artsy.net/admin/stages/141/deploy_strategies/37/edit) with arguments including
`"merge_after": <about 70 minutes from now in seconds>", "slack_webhook_url": "<my redacted test webhook>"` and confirm that a notification is sent. Then we can create a new webhook targeting #platform-machines (or #product-platform?), unset the `warned_pull_request_url` in the arguments, and repeat.

PR #221, opened 2020-09-16
Describe how to add new projects and deploy blocks

joeyAghion:
Horizon has gained lots of new features over time, but I don't think we've documented how to _fully_ set up new projects yet.

Once this is in place, my hope is to update the [more general deployments playbook](https://github.com/artsy/README/blob/master/playbooks/deployments.md) to link to it.


starsirius:
👍  Thanks for documenting the steps!

PR #229, opened 2020-09-23
Add Git

icirellik:
Add git to the release docker iamge.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM ruby:2.6.6-alpine AS base
 RUN apk update && apk --no-cache --quiet add --update \
+    git \
     nodejs \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -58,7 +58,8 @@ EXPOSE 3000
 WORKDIR /app
 
 RUN apk update && apk --no-cache --quiet add --update \
-    dumb-init
+    dumb-init \
+    git
```
joeyAghion, line 62: Since `git` is a runtime dependency, I think this should just go above into the `base` stage (and removed from the `builder` stage).

I'm not that familiar w/these new multi-stage builds though so up to you.

joeyAghion:
Stealing this back...

PR #247, opened 2020-10-17
Trying to make these a clearer.

pvinis:
While setting up Echo here https://releases.artsy.net/admin/stages/159/deploy_strategies/43, I noticed that I had no idea what `base` and `head` meant, and since they are not both branch names or both stage names, I thought it's better to document them.

For echo specifically, while I'm trying things out, I have `s3` branch (which eventually will be `master`) and `s3-production` (which eventually will be `production`).

With `staging` and `production` stages set up, both with the right branches, the deploy strategy needed `{"base"=>"s3-production", "head"=>"staging"}` to work, which seems unintuitive.


joeyAghion:
BTW these labels were based on [Github's API parameters](https://developer.github.com/v3/pulls/#input). (I also find them confusing.)

pvinis:
Oh I had no idea. Thanks for pointing that out.

PR #258, opened 2020-10-31
Fix hokusai dev.

artsyjian:
Bumped into problems upon `hokusai dev start`.

- App container exits. That's because Dockerfile `builder` target has no CMD. Add CMD to run Puma. Alternative is to add `command` in development.yml. Do we have a preference?

- Browser reports error:
```
Showing /app/app/views/projects/index.html.erb where line #3 raised:

Webpacker can't find projects.js in /app/public/packs/manifest.json. Possible causes:
...
```
That's b/c `/app/public` in the container is masked by the repo's `public` dir which is mounted as part of the bind mount `../:/app`.

Not sure about the best solution. Here I create a `named volume` for the dir. At the first start of the stack, Docker creates the volume and fills it with content from the container. The volume's existence causes the dir to be omitted from bind mount. Subsequent starts of the stack will use whatever's in the volume, which might become stale and require manual deletion.

Also create a named volume for `node_modules` for same reason.

```
$ docker volume ls|grep horizon
local               hokusai_horizon-node-modules
local               hokusai_horizon-public
```
- Populate container env with env files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,8 @@ COPY --chown=deploy:deploy . ./
 # Precompile Rails assets
 RUN bundle exec rake assets:precompile
 
+CMD ["bundle", "exec", "puma", "-C", "config/puma.rb"]
+
 FROM base AS production
 ENV PORT 3000
 EXPOSE 3000
--- a/hokusai/development.yml
+++ b/hokusai/development.yml
@@ -1,5 +1,10 @@
 ---
 version: "3.8"
+
+volumes:
+  horizon-node-modules:
+  horizon-public:
+
 services:
   horizon:
     build:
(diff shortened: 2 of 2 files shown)


izakp (approved):
I'm not sure what the best approach for `/app/public` is here either but if this is working for dev 👍

joeyAghion (commented):
I'm confused about the `public` path collision too. This project uses the `webpacker` gem which is designed to be compatible with rails' default usage of the `public` directory, so I suspect there's something atypical about Horizon's configuration causing the collision. I looked at a few of our examples ([Currents](https://github.com/artsy/currents/blob/2b3f6cf2395c4942c3ab9598b257348a34430bed/config/webpacker.yml), [Rosalind](https://github.com/artsy/rosalind/blob/fb4ac4f5ebe79b3469e3c385bf8515f0841ca81d/config/webpacker.yml), and [Volt](https://github.com/artsy/volt/blob/f7674d6ee6a4d114c6125c78a3aa6a3e0f340616/config/webpacker.yml)), and they don't specify `public_root_path: public` the way Horizon does, so maybe that's a clue.

artsyjian:
I get the same error from Volt started with hokusai dev start.
Screenshot from 2020-11-02 17-00-36

PR #301, opened 2021-01-06
Troubleshoot prod refresh comparison cron alert.

artsyjian:
https://artsy.slack.com/archives/CA8SANW3W/p1609878932042300

In troubleshooting the alerts (job failed status) we noticed in the logs that many of those cron's pods are sigterm'd. At least some of the sigterms are from cluster autoscaler (of background node group).

Adding `safe-to-evict: false` annotation to cron spec, which will cause CA to skip (for scale down) a node that has such a pod.

I am not sure that this will actually reduce the alerts, b/c it seems a pod being sigterm'd does not necessarily cause its parent `job` to be marked as failed (there's been more instances of sigterms than alerts).

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -144,6 +144,9 @@ spec:
   jobTemplate:
     spec:
       template:
+        metadata:
+          annotations:
+            "cluster-autoscaler.kubernetes.io/safe-to-evict": "false"
         spec:
           containers:
             - name: horizon-refresh-comparisons-cron


hokusai/production.yml
```diff
@@ -144,11 +144,14 @@ spec:
...
       template:
+        metadata:
+          annotations:
+            "cluster-autoscaler.kubernetes.io/safe-to-evict": "false"
         spec:
           containers:
             - name: horizon-refresh-comparisons-cron
               image: 585031190124.dkr.ecr.us-east-1.amazonaws.com/horizon:production
-              args: ["bundle", "exec", "rake", "cron:refresh_comparisons"]
+              command: ["bundle", "exec", "rake", "cron:refresh_comparisons"]
```
artsyjian, line 154: to override `ENTRYPOINT` (dumb-init) of docker image.

artsyjian:
reverted `args` -> `command` change.

izakp:
> At least some of the sigterms are from cluster autoscaler (of background node group).

Nice catch - we should definitely adopt this pattern elsewhere for non-idempotent crons

PR #315, opened 2021-01-26
fix: switch to python3/pip3

artsyjian:
pip doesn't work with python2 anymore.
since hokusai supports python3, switch to that and pip3.

https://app.circleci.com/pipelines/github/artsy/horizon/1298/workflows/883fe9ad-c4ae-4200-80fb-1ec2505843f7/jobs/1222

https://artsy.slack.com/archives/CA8SANW3W/p1611696568118800


izakp (approved):
:crossed_fingers:

izakp:
If we need to include https://github.com/artsy/hokusai/pull/247 fixes we can use a beta build

PR #320, opened 2021-02-04
fix: use hokusai beta

artsyjian:
Follow up to https://github.com/artsy/horizon/pull/315 in which we switched from python2/pip2 to python3/pip3.

Encountered a problem with hokusai registry images command under python3.

https://artsy.slack.com/archives/CA8SANW3W/p1611944044169800

We decided to switch horizon to hokusai beta release that has the fix, rather than waiting for an official release.

However, the beta Linux release is built on Debian (glibc) and does not work with Alpine (musl) which Horizon is based off.

https://artsy.slack.com/archives/CA8SANW3W/p1612374384229200

I attempted to build hokusai on alpine but encountered problem with Pyinstaller.

Therefore, for now, suggesting to install hokusai from Git (master). The install breaks with pip installed by apk. So install python/pip from source, and go with python3.5 since that's what hokusai officially supports.

Expecting builds to be slower and image size larger.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,15 @@ RUN apk update && apk --no-cache --quiet add --update \
 # horizon needs to compare production/staging envs of projects
 RUN apk update && apk --no-cache --quiet add --update \
     build-base \
-    python3-dev \
-    python3 \
-    py3-pip
+    openssl-dev
+
+RUN cd /tmp && \
+    wget -q https://www.python.org/ftp/python/3.5.8/Python-3.5.8.tgz && \
+    tar -xzf Python-3.5.8.tgz && \
+    cd Python-3.5.8 && \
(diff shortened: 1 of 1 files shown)


joeyAghion (approved):
Thanks!

izakp:
Thanks @artsyjian !  As far as the PyInstaller failures on Alpine go... they are distributed for convenience and so would not worry too much about supporting them in image builds.  Curious why the pip installed by apk failed though.

izakp:
For posterity:
```
Traceback (most recent call last):                           
  File "/usr/bin/pip", line 11, in <module>                                                       
    load_entry_point('pip==20.1.1', 'console_scripts', 'pip')() 
...
ModuleNotFoundError: No module named 'pip._vendor.packaging.tags'
```

PR #357, opened 2021-03-17
feat: protect cronjob from eviction by cluster-autoscaler, and misc.

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3197

- Set [safe-to-evict](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) to false for crons so they cannot be evicted by cluster-autoscaler.
- Set [backoffLimit](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy) for crons to reduce retries.
- Adjust resources to reflect usage.
- Generally, true up with hokusai [template](https://github.com/artsy/artsy-hokusai-templates/tree/master/rails-puma/hokusai).

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -32,7 +32,21 @@ spec:
           env:
             - name: PORT
               value: "8080"
-            - name: TRACE_AGENT_HOSTNAME
+            - name: PUMA_WORKERS
+              value: "1"
+            - name: PUMA_THREAD_MIN
+              value: "5"
+            - name: PUMA_THREAD_MAX
+              value: "5"
+            - name: RAILS_SERVE_STATIC_FILES
+              value: 'true'
+            - name: RAILS_LOG_TO_STDOUT
+              value: 'true'
+            - name: PUMA_BIND
+              value: 'tcp://0.0.0.0:8080'
+            - name: MALLOC_ARENA_MAX
+              value: "2"
+            - name: DATADOG_TRACE_AGENT_HOSTNAME
               valueFrom:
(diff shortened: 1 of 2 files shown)


hokusai/production.yml
```diff
@@ -60,11 +74,11 @@ spec:
             initialDelaySeconds: 10
             periodSeconds: 10
             timeoutSeconds: 5
-          lifecycle:
-            preStop:
-              exec:
-                command: ["sh", "-c", "sleep 10"]
-      dnsPolicy: Default
+      dnsPolicy: ClusterFirst
```
izakp, line 77: 😮 surprised we missed this here

izakp (approved):

hokusai/production.yml
```diff
@@ -32,7 +32,21 @@ spec:
...
+              value: "5"
+            - name: RAILS_SERVE_STATIC_FILES
+              value: 'true'
+            - name: RAILS_LOG_TO_STDOUT
+              value: 'true'
+            - name: PUMA_BIND
+              value: 'tcp://0.0.0.0:8080'
+            - name: MALLOC_ARENA_MAX
+              value: "2"
+            - name: DATADOG_TRACE_AGENT_HOSTNAME
```
joeyAghion, line 49: Surprised to see that this is unused. (Adding datadog can be a separate follow-up certainly.)

joeyAghion (approved):

PR #379, opened 2021-05-05
clean: pip install hokusai normally.

artsyjian:
revert https://github.com/artsy/horizon/pull/320 which was a work-around. hokusai 0.5.14 released with the fix so all good now.


joeyAghion:
Nice!

PR #404, opened 2021-06-16
Fix: gracefully ignore unverified commits when choosing deploy assignee

joeyAghion:
Horizon has been choking during its assignee-selection step, resulting in PRs without assignees and stale release data.

It turns out both `committer` and `author` can be `nil` on commit objects when "unverified." We hadn't encountered this until recently upon integrating external contractors. I'm not sure _why_ some commits are unverified or how to fix, but it's simple enough to exclude those commits from assignee logic.


jonallured (approved):
LGTM!! 🎉

PR #423, opened 2021-08-23
feat: allow comparison working dir to be specified via env.

artsyjian:
Supports https://artsyproduct.atlassian.net/browse/PLATFORM-3487

`perform_comparison` currently instantiates `Releasecop::Checker` with a new temp dir every time. `Releasecop::Checker` clones the project-at-task into that temp dir. The effect is that [refresh comparisons cron](https://github.com/artsy/horizon/blob/7d12644694e49f8cec5dfc97cfba9ad4122cb94e/Rakefile#L12) which runs every 10min clones all projects afresh at every run.

This PR provides the option for `perform_comparison` to use a fixed dir specified by the user via ENV.

The plan is to persist projects' clones in an EBS volume which will be mounted to the cron's pod on that fixed dir at every run. Thus, the cron clones afresh on the first run, and on subsequent runs, it only fetches deltas.

I have confirmed locally that [releasecop's cloning a project](https://github.com/joeyAghion/releasecop/blob/bc9c108f1c4ff33fefe75ffb0c44f5ac8822594d/lib/releasecop/checker.rb#L13) does not error if the clone already exists.

Merging this PR should not change the cron's behavior. If everything is good, we can proceed with adding an EBS volume and setting the fixed dir ENV, likely in `hokusai/staging.yml` and `hokusai/production.yml`.

### DIFF
--- a/app/services/releasecop_service.rb
+++ b/app/services/releasecop_service.rb
@@ -16,11 +16,11 @@ class ReleasecopService
   end
 
   def perform_comparison
-    env_dir = Horizon.config[:perform_comparison_workdir]
+    working_dir = Horizon.config[:working_dir]
 
-    if env_dir
-      Dir.mkdir(env_dir) unless Dir.exist?(env_dir)
-      perform_comparison_in_dir(env_dir)
+    if working_dir
+      Dir.mkdir(working_dir) unless Dir.exist?(working_dir)
+      perform_comparison_in_dir(working_dir)
     else
       Dir.mktmpdir(['releasecop', project.name]) do |dir|
         perform_comparison_in_dir(dir)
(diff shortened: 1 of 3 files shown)


app/services/releasecop_service.rb
```diff
@@ -16,13 +16,15 @@ def initialize(project)
...
 
   def perform_comparison
-    Dir.mktmpdir(['releasecop', project.name]) do |dir|
-      checker = Releasecop::Checker.new(
-        project.name,
-        project.stages.order(position: :asc).map { |s| build_manifest_item(s) },
-        dir
-      )
-      ResultWrapper.new(checker) # build comparisons
+    env_dir = Horizon.config[:perform_comparison_workdir]
```
joeyAghion, line 19: Minor: I'd call this variable and config `working_dir` for consistency and to match [releasecop](https://github.com/joeyAghion/releasecop/blob/bc9c108f1c4ff33fefe75ffb0c44f5ac8822594d/lib/releasecop/checker.rb#L5).

joeyAghion (approved):

app/services/releasecop_service.rb
artsyjian: sounds fine to me. updated.

PR #453, opened 2021-11-08
fea: Add blocked time period support

sonecabr:
**Description:**
As part of our automation we want to make our automated Deploy PR process as much flexible and reliable as possible, so we are introducing a new attribute into release automation called blocked_time_buckets which accepts an array of cron expression strings which can refine our automation to not create deploy PR's which are configured to release on blocked timespans. 
More info [here](https://artsyproduct.atlassian.net/browse/PLATFORM-3683)

**Compatibility:**
This config is backwards compatible. 
This config can be enabled by add a new attribute `blocked_time_buckets` into release config json
e.g.
https://releases.artsy.net/admin/stages/26/deploy_strategies/20 -> ARGUMENTS
from:
```
{"base"=>"release", "head"=>"staging"}
```
to:
```
{"base"=>"release", "head"=>"staging", "blocked_time_buckets": ["* 1-6 * * *", "* 18-22 * * *"]}
```
The expected behavior is that automated PR's will be only created on the time which fits into 7-18 and 23-00 daily.

### DIFF
--- a/app/services/deploy_service.rb
+++ b/app/services/deploy_service.rb
@@ -121,12 +121,11 @@ class DeployService
       .detect { |l| github_client.check_assignee(github_repo, l) }
   end
 
-  def can_release_now?(blocked_buckets, reference_time)
-    return if !blocked_buckets || !reference_time
+  def can_release_now?(buckets, reference_time)
+    return if !buckets || !reference_time
 
     time_wa = reference_time.beginning_of_minute.strftime('%a, %d %b %Y %H:%M')
-    (raise "Merge time is blocked") 
-      if blocked_buckets.map { |blocked_bucket| cron_match(time_wa, blocked_bucket) }.reduce(:|)
+    buckets.map { |blocked_bucket| cron_match(time_wa, blocked_bucket) }.reduce(:|) || (raise 'Merge time blocked')
   end
 
   def cron_match?(reference_time, blocked_bucket)
--- a/app/services/deploy_service.rb
+++ b/app/services/deploy_service.rb
@@ -125,6 +125,8 @@ class DeployService
     return if !buckets || buckets.empty? || !reference_time
 
     time_wa = reference_time.beginning_of_minute.strftime('%a, %d %b %Y %H:%M')
+    pp time_wa
+    pp buckets
     buckets.map { |blocked_bucket| cron_match?(time_wa, blocked_bucket) }.reduce(:|) || (raise 'Merge time blocked')
   end
 
--- a/app/services/deploy_service.rb
+++ b/app/services/deploy_service.rb
@@ -21,17 +21,17 @@ class DeployService
   private
 
   def create_github_pull_request
-    can_release_now?(deploy_strategy.arguments['blocked_time_buckets'], Time.now)
-    pull_request = github_client.create_pull_request(
-      github_repo,
-      deploy_strategy.arguments['base'],
-      deploy_strategy.arguments['head'],
-      'Deploy',
-      'This is an automatically generated release PR!'
-    )
-
-    assignee = choose_assignee(pull_request)
-    github_client.add_assignees(github_repo, pull_request.number, assignee) if assignee
+    if can_release_now?(deploy_strategy.arguments['blocked_time_buckets'] || [], Time.now)
+      pull_request = github_client.create_pull_request(
+        github_repo,
+        deploy_strategy.arguments['base'],
+        deploy_strategy.arguments['head'],
+        'Deploy',
+        'This is an automatically generated release PR!'
+      )
+      assignee = choose_assignee(pull_request)
+      github_client.add_assignees(github_repo, pull_request.number, assignee) if assignee
+    end
   rescue Octokit::UnprocessableEntity # release PR already exists
     pull_request = github_client.pull_requests(
       github_repo,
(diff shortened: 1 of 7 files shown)


app/services/deploy_service.rb
```diff
@@ -48,6 +48,7 @@ def create_github_pull_request
           return
         end
       end
+      can_release_now?(deploy_strategy.arguments['blocked_time_buckets'], merge_at)
```
joeyAghion, line 51: This comes a few lines _after_ the release PR is merged. It should be checked _before_ merging and ideally before creating new release PRs.

Also, shouldn't `Time.now` be used rather than `merge_at` (which is some past time)?

app/services/deploy_service.rb
```diff
@@ -119,4 +120,16 @@ def choose_assignee(pull_request)
...
       .detect { |l| github_client.check_assignee(github_repo, l) }
   end
+
+  def can_release_now?(buckets, reference_time)
+    buckets = Array(buckets)
+    return if !buckets || buckets.empty? || !reference_time
+
+    time_wa = reference_time.beginning_of_minute.strftime('%a, %d %b %Y %H:%M')
+    buckets.map { |blocked_bucket| cron_match?(time_wa, blocked_bucket) }.reduce(:|) || (raise 'Merge time blocked')
+  end
```
joeyAghion, line 130: This helper should return `true`/`false` rather than raising an exception. The `any?`/`none?` helpers can save a little time by returning as soon as one of the conditions is true.

```ruby
buckets = deploy_strategy.arguments['blocked_time_buckets'] || []
buckets.none? { |bucket| Fugit::Cron.parse(bucket).match?(Time.now) }
```

app/services/deploy_service.rb
sonecabr: Hi @joeyAghion , I used merge_at actually because I assumed that we where more interested in the time this PR is planned to be merged, I will change it to use now instead.
I will also put the check before the merge, I think I mixed the version somehow, probably doing some rollback on the tests I accidentally put it back on the wrong place.

sonecabr: Just pushed the changes :)

app/services/deploy_service.rb
sonecabr: I also had a version returning true or false, but I have changed to throw an exception after noticed another helper which checks the authentication trowing an error during tests implementation. I will bring back the true/false idea.

sonecabr: Just updated it @joeyAghion , hope it matches to what you suggested.

app/services/deploy_service.rb
```diff
@@ -119,4 +120,9 @@ def choose_assignee(pull_request)
       .group_by(&:login).map { |k, v| [v.size, k] }.sort.reverse.map(&:last)
       .detect { |l| github_client.check_assignee(github_repo, l) }
   end
+
+  def can_release_now?(buckets, reference_time)
+    time_wa = reference_time.beginning_of_minute.strftime('%a, %d %b %Y %H:%M')
```
joeyAghion, line 125: Why convert to a string here?

Bonus: you can provide a default argument of `reference_time = Time.now` so that callers don't have to pass the 2nd argument (unless they choose to).

sonecabr: just added the default, I used to have defaults on python...completely forgot that ruby also supports it.

spec/features/deploy_spec.rb
```diff
@@ -315,4 +316,14 @@
       expect(webhook).not_to have_been_made
     end
   end
+
+  it 'failed to release due blocked time' do
+    strategy.update!(arguments: strategy.arguments.merge(
+      blocked_time_buckets: ['* 1-23 * * *'],
```
joeyAghion, line 322: Should this be `0-23`? Otherwise it will "fail" (allow a deploy) when tests are run at certain hours of the night.

app/services/deploy_service.rb
```diff
@@ -21,16 +21,17 @@ def start
...
+      pull_request = github_client.create_pull_request(
+        github_repo,
+        deploy_strategy.arguments['base'],
+        deploy_strategy.arguments['head'],
+        'Deploy',
+        'This is an automatically generated release PR!'
+      )
+      assignee = choose_assignee(pull_request)
+      github_client.add_assignees(github_repo, pull_request.number, assignee) if assignee
+    end
```
joeyAghion, line 34: I think this works, but I'm noticing that the logic has become complex. There are all of these code locations testing whether the deploy can proceed:

https://github.com/artsy/horizon/blob/54d9646244b1ef9aae60184fa4cb607ed71a3742/app/services/comparison_service.rb#L64

And (here):

```
if can_release_now?(deploy_strategy.arguments['blocked_time_buckets'] || [])
```

And: 

https://github.com/artsy/horizon/blob/54d9646244b1ef9aae60184fa4cb607ed71a3742/app/services/deploy_service.rb#L41-L46

In a later PR, maybe we should refactor these into a single condition that prevents the whole `DeployService...` from being invoked.

joeyAghion (approved):

spec/features/deploy_spec.rb
sonecabr: yup, just changed.

app/services/deploy_service.rb
sonecabr: Indeed, I think I will be able to do a refactor near in future once I get more familiar with horizon.

PR #465, opened 2021-12-03
chore: bump webpacker 5.4.3

mc-jones:
This PR updates webpacker to the latest stable version to address security vulnerabilities in 5.1.1.

- `tar` (high severity): https://github.com/artsy/horizon/security/dependabot/yarn.lock/tar/open
- `serialize-javascript` (high severity): https://github.com/artsy/horizon/security/dependabot/yarn.lock/serialize-javascript/open
- `trim-newlines` (high severity): https://github.com/artsy/horizon/security/dependabot/yarn.lock/trim-newlines/open

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -99,7 +99,7 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.12.2)
-    concurrent-ruby (1.1.8)
+    concurrent-ruby (1.1.9)
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     crass (1.0.6)
@@ -123,7 +123,7 @@ GEM
       actionpack (>= 5.2)
       activesupport (>= 5.2)
     hashdiff (1.0.1)
-    i18n (1.8.10)
+    i18n (1.8.11)
       concurrent-ruby (~> 1.0)
     inherited_resources (1.12.0)
       actionpack (>= 5.2, < 6.2)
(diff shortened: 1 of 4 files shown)


babel.config.js
```diff
@@ -27,6 +27,18 @@ module.exports = function (api) {
...
+        {
+          loose: true,
+        },
+      ],
+      [
+        "@babel/plugin-proposal-private-property-in-object",
+        {
+          loose: true,
+        },
+      ],
```
mc-jones, line 41: `@babel/plugin-proposal-class-properties`, `@babel/plugin-proposal-private-methods` and `@babel/plugin-proposal-private-property-in-object` all need to have the same option for `loose`, per babel docs [here](babel/plugin-proposal-private-methods#loose) and [here](https://babeljs.io/docs/en/babel-plugin-proposal-private-property-in-object#loose).

Gemfile.lock
```diff
@@ -311,7 +311,7 @@ GEM
       rails (>= 3.2.16)
     xpath (3.2.0)
       nokogiri (~> 1.8)
-    zeitwerk (2.4.2)
+    zeitwerk (2.5.1)
```
joeyAghion, line 314: Are these upgrades related at all?

joeyAghion (approved):

Gemfile.lock
mc-jones: They should be `webpacker` dependency updates as they were updated as a result of `bundle update webpacker`. We had updated the Gemfile directly when bumping just the Ruby side of webpacker previously, see [PR](https://github.com/artsy/horizon/commit/92db616ffb7f7f90686df18327a7df24b726e37c), which is why we aren't seeing a direct `webpacker` bump in this PR, but just its dependencies.

For this update, I followed the full upgrade instructions from `webpacker` [here](https://github.com/rails/webpacker#upgrading).

PR #467, opened 2021-12-07
git remote of project can still be found when default branch is main

kajatiger:
Connected to https://github.com/artsy/exchange/pull/1008 and to [PX-4258](https://artsyproduct.atlassian.net/browse/PX-4258)

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -28,7 +28,7 @@ class Project < ApplicationRecord
   end
 
   def git_remote
-    stage = stages&.detect { |s| s.name == 'master' }
+    stage = stages&.detect { |main_stage| %w[main master].any? main_stage.name }
     stage&.git_remote
   end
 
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -28,7 +28,7 @@ def github_repo_url
   end
 
   def git_remote
-    stage = stages&.detect { |s| s.name == 'master' }
+    stage = stages&.detect { |main_stage| %w[main master].any? main_stage.name }
```
joeyAghion, line 31: I didn't know about this style of `any?`!

joeyAghion (approved):

app/models/project.rb
kajatiger: I honestly only prefer it because I am getting `include?` 50% of the time wrong with adding an `s` for the third person flexion. 🙈

PR #503, opened 2022-05-16
chore: update rails to 6.1.5.1

mc-jones:
This PR bumps rails from `6.1.3.1` to `6.1.5.1`. Motivation is to bump `actionpack`. Tested locally and it ran fine, but there is a limit to what you can test locally with `horizon` so will announce in #dev for people to let us know if something looks odd.


artsyjian (approved):
lgtm

PR #506, opened 2022-05-18
chore(deps): update webpack-bundle-analyzer

kajatiger:
Dependabot cannot update ejs to a non-vulnerable version

The latest possible version that can be installed is 2.7.4 because of the following conflicting dependency:

webpack-bundle-analyzer@3.8.0 requires ejs@^2.6.1

The earliest fixed version is 3.1.7.


mc-jones:
👍 🙇

mc-jones (approved):

PR #528, opened 2022-07-27
chore: bump rails to 6.1.6.1

mc-jones:
This PR bumps rails to `6.1.6.1` from `6.1.5.1`


joeyAghion (approved):

kajatiger:
I will fetch the branch and try out with this farady-retry gem. Maybe this can fix the problem.

kajatiger:
So adding the faraday-retry did not fix it, but adding coffee-rails did. After that there were some rubocop offenses to fix for the CI to run the tests.

mc-jones:
Thanks @kajatiger! 

Quick question for your ruby expertise: it looks like the only reason we need `coffee-rails` is for [this single](https://github.com/artsy/horizon/blob/main/vendor/assets/javascripts/active_admin.js.coffee) coffee script file, which looks to just be commented out. Seems this was possibly part of earlier rails versions? Do you think it's fine to delete and remove the need for the dependency altogether?

kajatiger:
@mc-jones if we really don't need coffee-script anywhere in this repo then I think your suggestion would be an awesome solution!

mc-jones:
@kajatiger 👍. I'm going to merge this to get the update in and can look at the dependency separately.

PR #558, opened 2023-02-08
POC - Rails - standard logging format

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4904

This PR introduces a new logger [lograge](https://github.com/roidrage/lograge) (replacing default rails logger) for the purposes of proving out a POC which is about _standardizing logging posture across Artsy applications_ and specifically as it pertains to web apps.

<details>
<summary>The following options have been evaluated before moving forward with lograge</summary>

</details>

Choosing _lograge_ seemed like a no-brainer seeing as we already import this gem in [a couple of projects](https://github.com/search?q=org%3Aartsy+lograge&type=code) and among alternatives: this gem is quite popular, well documented and still maintained.

### default logger vs lograge

> examples pulled from production mode

```ruby
# default
I, [2023-02-08T10:53:34.520694 #12855]  INFO -- : [27a4c934-f255-4336-ab11-205323e6c4b3] Started GET "/admin/projects" for 127.0.0.1 at 2023-02-08 10:53:34 -0500
I, [2023-02-08T10:53:34.521270 #12855]  INFO -- : [27a4c934-f255-4336-ab11-205323e6c4b3] Processing by Admin::ProjectsController#index as HTML
I, [2023-02-08T10:53:34.575644 #12855]  INFO -- : [27a4c934-f255-4336-ab11-205323e6c4b3]   Rendered /Users/[REDACTED]/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/activeadmin-2.13.1/app/views/active_admin/resource/index.html.arb (Duration: 40.0ms | Allocations: 32988)
I, [2023-02-08T10:53:34.575858 #12855]  INFO -- : [27a4c934-f255-4336-ab11-205323e6c4b3] Completed 200 OK in 55ms (Views: 38.6ms | ActiveRecord: 1.9ms | Allocations: 33562)
I, [2023-02-08T10:53:51.645002 #12855]  INFO -- : [7f51bc38-d881-4534-b361-5e5f9d0ffcf5] Started GET "/admin/profiles" for 127.0.0.1 at 2023-02-08 10:53:51 -0500
I, [2023-02-08T10:53:51.646074 #12855]  INFO -- : [7f51bc38-d881-4534-b361-5e5f9d0ffcf5] Processing by Admin::ProfilesController#index as HTML
[...]

PR #560, opened 2023-02-08
feat: use safe navigation when accessing request_id in lograge custom_options

ovasdi:
Followup to https://github.com/artsy/horizon/pull/558

Add safe navigation when setting `request_id` in lograge `custom_options`. This prevents `action_cable` errors observed when accessing the `/` route.

```
Could not log "connect.action_cable" event. NoMethodError: undefined method `[]' for nil:NilClass
...
```

### DIFF
--- a/config/initializers/lograge.rb
+++ b/config/initializers/lograge.rb
@@ -5,7 +5,7 @@ Rails.application.configure do
   config.lograge.formatter = Lograge::Formatters::Json.new
   config.lograge.custom_options = lambda do |event|
     {
-      request_id: event.payload[:headers]['action_dispatch.request_id']
+      request_id: event.payload[:headers]&.[]('action_dispatch.request_id')
     }.compact
   end
 end


joeyAghion (approved):

config/initializers/lograge.rb
```diff
@@ -5,7 +5,7 @@
   config.lograge.formatter = Lograge::Formatters::Json.new
   config.lograge.custom_options = lambda do |event|
     {
-      request_id: event.payload[:headers]['action_dispatch.request_id']
+      request_id: event.payload[:headers]&.[]('action_dispatch.request_id')
```
ovasdi, line 8: As this wasn't explained in code to describe this elegant approach to check that a key exists in `payload`, leaving this breadcrumb in the PR for those who may look at this line in the future and in case it isn't clear what this is.

The `&.` is known as [Safe Navigation Operator](https://ruby-doc.org/core-3.0.2/doc/syntax/calling_methods_rdoc.html#label-Safe+Navigation+Operator) and combined with [] it is equivalent of using `hash && hash[:key]` which allows safely navigating the `payload` which may not always contain the `headers` key.

PR #562, opened 2023-02-09
fix: failing ruby upgrade because of keyword argument syntax change

kajatiger:
An untested service failed with the new ruby version. More info here: https://artsy.slack.com/archives/CA8SANW3W/p1675958384294309

This is a fix
To-Do: write another test for the comparison service.

### DIFF
--- a/spec/services/comparison_service_spec.rb
+++ b/spec/services/comparison_service_spec.rb
@@ -5,7 +5,7 @@ require 'rails_helper'
 RSpec.describe ComparisonService, type: :model do
   include ActiveSupport::Testing::TimeHelpers
   let(:org) { Organization.create! name: 'Artsy' }
-  let(:project) do
+  let!(:project) do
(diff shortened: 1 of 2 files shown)


spec/services/comparison_service_spec.rb
```diff
@@ -14,12 +14,18 @@
...
+      allow_any_instance_of(ComparisonService).to receive(:refresh_comparisons)
+        .and_raise(ZeroDivisionError)
       expect { ComparisonService.refresh_all_comparisons }.not_to raise_error
     end
   end
 
+  describe 'refresh_comparisons_for_organizations' do
+    it 'works' do
+      expect { ComparisonService.refresh_comparisons_for_organization(org) }.not_to raise_error
+    end
```
kajatiger, line 26: This one actually calls the action cable broadcast and throws the exact same error that we saw on papertrail.

PR #579, opened 2023-06-26
chore: add DogStatsD

mc-jones:
This PR adds DogStatsD (via Datadog's [ruby gem](https://github.com/DataDog/dogstatsd-ruby)) to `horizon`, which will allow us to send custom metrics to our Datadog instance. This should be generally useful, but it is being added now to[ support better monitor and tracking of our system runtimes versions](https://artsyproduct.atlassian.net/browse/PLATFORM-5073). 

There will be a follow-up PR to instrument implement runtime version monitoring, but I wanted to keep the PRs focused and separate.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ ruby File.read('.ruby-version')
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'coffee-rails'
+gem 'dogstatsd-ruby', require: 'datadog/statsd' # send metrics to datadog agent
 gem 'rails', '~> 6.1.6.1'
 # Use postgresql as the database for Active Record
 gem 'pg'
--- a/config/initializers/_config.rb
+++ b/config/initializers/_config.rb
@@ -9,7 +9,7 @@ Horizon.config = {
   minimum_version_node: ENV.fetch('MINIMUM_VERSION_NODE', nil),
   working_dir: ENV.fetch('WORKING_DIR', nil),
   datadog_service_name: ENV['DATADOG_SERVICE_NAME'] || 'horizon',
-  datadog_trace_agent_hostname: ENV.fetch('DATADOG_TRACE_AGENT_HOSTNAME', 'datadog')
+  datadog_trace_agent_hostname: ENV.fetch('DATADOG_TRACE_AGENT_HOSTNAME', nil)
 }
 
 if Rails.env.production? && Horizon.config[:basic_auth_pass].blank?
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -7,6 +7,7 @@ ruby File.read('.ruby-version')
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'coffee-rails'
+gem 'dogstatsd-ruby', require: 'datadog/statsd' # send metrics to datadog agent
```
mc-jones, line 10: I thought there might be a peer dependency on `ddtrace` but while testing locally I was able to report metrics just with `dogstatsd-ruby`.

config/initializers/_config.rb
```diff
@@ -7,7 +7,9 @@
   basic_auth_pass: ENV.fetch('BASIC_AUTH_PASS', nil),
   minimum_version_ruby: ENV.fetch('MINIMUM_VERSION_RUBY', nil),
   minimum_version_node: ENV.fetch('MINIMUM_VERSION_NODE', nil),
-  working_dir: ENV.fetch('WORKING_DIR', nil)
+  working_dir: ENV.fetch('WORKING_DIR', nil),
+  datadog_service_name: ENV['DATADOG_SERVICE_NAME'] || 'horizon',
+  datadog_trace_agent_hostname: ENV.fetch('DATADOG_TRACE_AGENT_HOSTNAME', nil)
```
mc-jones, line 12: Will be set via [k8](https://github.com/artsy/horizon/blob/main/hokusai/production.yml#L50)

config/initializers/dogstatsd.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Horizon
+  def self.dogstatsd
+    @dogstatsd ||= Datadog::Statsd.new(
+      Horizon.config[:datadog_trace_agent_hostname],
+      8125,
```
mc-jones, line 7: @ovasdi - Calling this out since we were confused, but the `datadog` agent does listen for this on port `8125` per the [docs](https://docs.datadoghq.com/developers/dogstatsd/?tab=hostagent#setup).

Gemfile
joeyAghion: Can't believe that Horizon never had `ddtrace`! :face-palm:

joeyAghion (approved):

config/initializers/dogstatsd.rb
ovasdi: Sorry for throwing you off with some of my suggestions. It seems hooking this up was much simpler in the end.

PR #580, opened 2023-06-27
chore: use docker buildkit

mc-jones:
This PR updated the `.circleci/config.yml` file to use docker buildkit. For more context and motivation please see the Slack [thread](https://artsy.slack.com/archives/CA8SANW3W/p1687874328399399). 

### Motivation
While merging an unrelated PR, we ran into an issue where the production image of our multi-stage build was not being pushed to our registry, due to having the same git SHA, which we use as the image tag. 

### Solution
By utilizing our existing [artsy-remote-docker/buildkit-*](https://github.com/artsy/orbs/blob/7c26a69a6b560d6acb4b31fb267244d89ea42d6a/src/remote-docker/remote-docker.yml#L256) steps, we're able to leverage docker's [buildkit](https://docs.docker.com/build/buildkit/), as well as benefit from the auto appending of the stage to the image tag. For example:
```
horizon:main: hokusai registry images
Image Pushed At           | Image Tags
----------------------------------------------------------
2023-06-27 12:03:50-04:00 | d98685f91ae9bf3886791e3c3b5020920c55a77c-builder
2023-06-27 08:23:55-04:00 | b615a89d8f25b7c594ae9a6f7935484187ae1576
2023-06-27 08:23:47-04:00 | 1601aae393f07ab514bbf2cc99e77965dc4f69e2
2023-06-27 08:23:42-04:00 | c57874f4e8d697e46ffd2867c5ec9c40b8daa6ff
2023-06-26 15:59:15-04:00 | 5d05480a01f4311c72647c5326ef0b15b19d8d21, 400cd47b2779dbe9b9fcc68dcbf865abbb5872b4
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
     executor: hokusai/deploy
     steps:
       - hokusai/setup-docker
-      - run: hokusai registry pull --tag "$CIRCLE_SHA1"
+      - run: hokusai registry pull --tag "$CIRCLE_SHA1"-builder
       - hokusai/run-tests:
           flags: --no-build
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,13 +42,13 @@ workflows:
           project_id: 38
 
       - artsy-remote-docker/buildkit-build:
-            <<: *not_staging_or_release
-            context: hokusai
-            executor: hokusai/beta
-            name: builder-image-build
-            pre-steps:
-              - run:
-                  command: echo 'export BUILD_TARGET="builder";' >> $BASH_ENV
+          <<: *not_staging_or_release
+          context: hokusai
+          executor: hokusai/beta
+          name: builder-image-build
+          pre-steps:
+            - run:
+                command: echo 'export BUILD_TARGET="builder";' >> $BASH_ENV
 
       - artsy-remote-docker/buildkit-push:
           <<: *not_staging_or_release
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -41,35 +41,54 @@ workflows:
           context: horizon
           project_id: 38
 
-      - artsy-remote-docker/build:
+      - artsy-remote-docker/buildkit-build:
           <<: *not_staging_or_release
           context: hokusai
-          name: build
+          executor: hokusai/beta
```
artsyjian, line 47: This in Force' case is for testing [Hokusai beta release](https://github.com/artsy/hokusai#beta-release). Most projects use [Hokusai canonical release](https://github.com/artsy/orbs/blob/7c26a69a6b560d6acb4b31fb267244d89ea42d6a/src/hokusai/hokusai.yml#L10). Can keep this if we want Horizon to also be a Hokusai-beta tester.

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     executor: hokusai/deploy
     steps:
       - hokusai/setup-docker
-      - run: hokusai registry pull --tag "$CIRCLE_SHA1"
+      - run: hokusai registry pull --tag "$CIRCLE_SHA1"-builder
```
artsyjian, line 13: :+1:

artsyjian (approved):
:100:

.circleci/config.yml
mc-jones: Good catch. I'll switch over to the canonical release. I don't think Horizon gets deployed enough to make it particularly useful as a canary, and could just confuse devs in the event it does introduce some breaking change.

mc-jones: Switched in [5d71a00](https://github.com/artsy/horizon/pull/580/commits/5d71a000979b343c3310b15c2b69e831927b0912)

artsyjian: Okay cool. Then the `executor` spec here can be omitted altogether (Orb defaults to canonical).

artsyjian: my fault, `executor` is required at one place - [the `test` job definition](https://github.com/artsy/horizon/blob/400cd47b2779dbe9b9fcc68dcbf865abbb5872b4/.circleci/config.yml#L10C1-L11C1).

PR #581, opened 2023-06-28
feat: send runtime version status to datadog

mc-jones:
### Problem 

We don't currently have a way of effectively informing team members when their project's runtime dependencies are out of date. While `horizon` does have a UI to display which projects need runtime updates, engineers do not seem to be [using it](https://artsy.slack.com/archives/C02BC3HEJ/p1687284843609379). 

### Solution

Since we are already updating project runtime status in `horizon`, we can also send this information to `datadog`, which will allow us to build dashboards, monitors, and trigger alerts. 

[PLATFORM-5073]

[PLATFORM-5073]: https://artsyproduct.atlassian.net/browse/PLATFORM-5073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/project_data_service.rb
+++ b/app/services/project_data_service.rb
@@ -56,19 +56,9 @@ class ProjectDataService
 
   def update_dependency(name, version)
     dependency = @project.dependencies.find_or_initialize_by(name: name)
-
     dependency.update(version: version)
 
-    Horizon.dogstatsd.gauge(
-      'runtime.version_status', # Metric name
-      dependency.update_required? ? -1 : 1, # The value associated with the metric. -1 = out of date, 1 = up to date
-      tags: [
-        "runtime:#{name}",
-        "project:#{@project.name}",
-        "criticality:#{@project.criticality}",
-        "team:#{@project.tags&.join(':')}"
-      ]
-    )
+    report_runtime_version_status(dependency)
 
     dependency
   end
@@ -120,5 +110,18 @@ class ProjectDataService
     file&.to_h
   rescue Octokit::NotFound
     # file not found - don't fail if ruby project doesn't have node etc
+  end
+
+  def report_runtime_version_status(dependency)
+    Horizon.dogstatsd.gauge(
+      'runtime.version_status', # Metric name
+      dependency.update_required? ? -1 : 1, # The value associated with the metric. -1 = out of date, 1 = up to date
+      tags: [
+        "runtime:#{dependency.name}",
+        "project:#{@project.name}",
+        "criticality:#{@project.criticality}",
+        "team:#{@project.tags&.join(':')}"
+      ]
+    )
   end
 end
--- a/app/services/project_data_service.rb
+++ b/app/services/project_data_service.rb
@@ -120,7 +120,7 @@ class ProjectDataService
         "runtime:#{dependency.name}",
         "project:#{@project.name}",
         "criticality:#{@project.criticality}",
-        "team:#{@project.tags&.join(':')}"
+        "team:#{@project.tags&.none? ? 'none' : @project.tags.join(':')}"
       ]
     )
   end
--- a/app/services/project_data_service.rb
+++ b/app/services/project_data_service.rb
@@ -60,9 +60,10 @@ class ProjectDataService
     dependency.update(version: version)
 
     Horizon.dogstatsd.gauge(
-      "runtime.#{name}.version_status", # Metric name
+      'runtime.version_status', # Metric name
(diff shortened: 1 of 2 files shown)


app/services/project_data_service.rb
```diff
@@ -107,4 +111,17 @@ def fetch_github_file(path)
...
+
+  def report_runtime_version_status(dependency)
+    Horizon.dogstatsd.gauge(
+      'runtime.version_status', # Metric name
+      dependency.update_required? ? -1 : 1, # The value associated with the metric. -1 = out of date, 1 = up to date
+      tags: [
+        "runtime:#{dependency.name}",
+        "project:#{@project.name}",
+        "criticality:#{@project.criticality}",
+        "team:#{@project.tags&.join(':')}"
```
mc-jones, line 123: We currently accept `nil` team values. This seems reasonable, though ends up looking kind of weird in datadog, with a tag of just `team:`. Happy to hear alternatives here.

app/services/project_data_service.rb
```diff
@@ -107,4 +111,17 @@ def fetch_github_file(path)
   rescue Octokit::NotFound
     # file not found - don't fail if ruby project doesn't have node etc
   end
+
+  def report_runtime_version_status(dependency)
+    Horizon.dogstatsd.gauge(
+      'runtime.version_status', # Metric name
+      dependency.update_required? ? -1 : 1, # The value associated with the metric. -1 = out of date, 1 = up to date
```
mc-jones, line 118: I'm not totally convinced that this is the best approach but think it is a reasonable start. I want to let this run in staging and play with different dashboard options and will report back to the team on what I think makes sense. I sent a few data points via a rails console but it proved difficult to make good monitors with so little data.

app/services/project_data_service.rb
mc-jones: I also wonder if the tag names are too generic and should be prepended with `horizon`?

ovasdi: > We currently accept nil team values. This seems reasonable, though ends up looking kind of weird in datadog, with a tag of just team:. Happy to hear alternatives here.

What if we do not include the `team:` tag in the list when a project has no team value? Or set `none` or something similar?

---

> I also wonder if the tag names are too generic and should be prepended with horizon?

So it would be?:
```
horizon-runtime:ruby
horizon-project:eigen
...
```

I suppose if it's not obvious that these are coming from Horizon it may be helpful but it doesn't seem necessary.

ovasdi (approved):
LGTM

app/services/project_data_service.rb
mc-jones: > Or set none or something similar?

Good call! I think this is probably the best option and matches what does when it tries to set a tag with an unknown value (e.g. `env:none` when developing locally). 

> I suppose if it's not obvious that these are coming from Horizon it may be helpful but it doesn't seem necessary.

Yeah, that's what I'm thinking. We will be able to filter by `service:horizon` to always filter by `horizon`.

Screenshot 2023-06-29 at 9 23 59 AM

I'll keep it as is for now and we can reconsider if it becomes confusing or problematic.

mc-jones:
Alright. Thanks for the feedback @ovasdi. Going to remove draft status.

app/services/project_data_service.rb
joeyAghion: I agree it's a reasonable start. If we could _reliably_ calculate the difference between target and actual versions, that would be nicer. Something like:

```ruby
target_version = Gem::Version.new(TARGET)
actual_version = Gem::Version.new(actual)
actual_version.segments.first - target_version.segments.first
```
(This would calculate the _major_ version difference, but maybe there's a way to be even more granular.)

app/services/project_data_service.rb
```diff
@@ -107,4 +111,17 @@ def fetch_github_file(path)
...
+      tags: [
+        "runtime:#{dependency.name}",
+        "project:#{@project.name}",
+        "criticality:#{@project.criticality}",
+        "team:#{@project.tags&.none? ? 'none' : @project.tags.join(':')}"
```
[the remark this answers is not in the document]
mc-jones: This was largely a response to checking the `tags` on horizon and seeing they were exclusively teams, though I agree it is better to give ourselves greater flexibility up front and narrow later if there seems a better approach.

I've updated this in #582.

PR #651, opened 2024-10-09
chore: Switch to asdf

jonallured:
For whatever reason this project had not yet switched to asdf so that's what this PR does.

/cc @artsy/amber-devs


jonallured:
Merging!

PR #665, opened 2024-11-14
feat: retrieve ruby version from Gemfile when available

joeyAghion:
I noticed [Horizon's detailed view of project runtimes and dependencies](https://releases.artsy.net/?view=list) wasn't aware of `asdf`, and continued to try to load ruby and nodejs versions from `.ruby-version` and `.nvmrc` files. Upon investigation, I found that it was already loading `Gemfile`s for any project without `.ruby-version`, so decided to just get the ruby version from that instead of from `.tool-versions`.

I considered doing away with the `.nvmrc` check as well, but the nodejs version from `package.json` already takes precedence, so should continue to work for `asdf` projects.

### DIFF
--- a/app/services/project_data_service.rb
+++ b/app/services/project_data_service.rb
@@ -3,6 +3,8 @@
 require "base64"
 
 class ProjectDataService
+  GEMFILE_RUBY_REGEX = /^ruby "([0-9.]+)"$/
+
   def initialize(project)
     @project = project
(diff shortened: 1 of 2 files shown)


jonallured:
Just wait 😝 

https://github.com/jonallured/monolithium/blob/main/Gemfile#L3

jonallured (approved):

joeyAghion:
Woah!

PR #670, opened 2024-11-27
feat: do not ignore load secrets script

artsyjian:
We have to not ignore the script.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -33,3 +33,6 @@
 !storage/.keep
 !tmp/.keep
 !vendor/
+
+# scripts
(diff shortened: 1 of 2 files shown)


ovasdi (approved):

.dockerignore
```diff
@@ -33,3 +33,6 @@
 !storage/.keep
 !tmp/.keep
 !vendor/
+
+# scripts
+!scripts/load_secrets_and_run.sh
```
ovasdi, line 38: Oh so this is why the build used the cache even though that dir/file was new🤔

PR #682, opened 2025-02-25
chore: update pr description

egdbear:
Since we can globally enable automatic PR merging, we can remove this text.


joeyAghion (approved):
🙏

PR #764, opened 2026-07-22
chore: impose 10d cooldown period before dependency updates

narikazu:
Mirrors [artsy/gravity#20262](https://github.com/artsy/gravity/pull/20262) — upgrades `bundler` for the ability to specify a `cooldown` argument, delaying dependency updates until they've had 10 days to be vetted elsewhere. This mitigates supply-chain risk from a compromised gem release being auto-picked-up before the community notices.

Requires Bundler >= 4.0.13 for the `cooldown` source option; this repo was still on the 2.x line, so this bumps bundler across a major version (2.4.5 → 4.0.17).

Assisted by Claude Code Sonnet 5

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,8 @@ USER deploy
 # TODO: look into buildkit to prevent re-installing node modules after changing gems
 COPY --chown=deploy:deploy Gemfile Gemfile.lock ./
 # RUN bundle install -j4 --path /usr/local/bundle-prod --without development test && \
-RUN bundle install -j4 --path /usr/local/bundle
+RUN bundle config set --global path /usr/local/bundle && \
+    bundle install -j4
 # bundle clean
 
 # Set up packages, empty cache to save space
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-source "https://rubygems.org"
+source "https://rubygems.org", cooldown: 10
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby "3.3.0"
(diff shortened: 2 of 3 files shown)


artsyjian (approved):
thanks!

narikazu:
Fixed the CI failure: the Dockerfile's `bundle install -j4 --path /usr/local/bundle` used the `--path` flag, which Bundler 4 removed entirely (it relied on being remembered across invocations, which Bundler no longer does). Replaced with `bundle config set --global path /usr/local/bundle` followed by a flag-less `bundle install -j4`. Verified locally against this repo's Gemfile/Gemfile.lock before pushing.

Assisted by Claude Code Sonnet 5

narikazu:
This is a real bug, not a fluke — it's baked into the built image itself, so it would fail on any deploy that runs the \`bundle exec rake db:migrate\` pre-deploy hook against this image (staging likely didn't hit it because it ran an older image or skipped the hook this time).

Root cause: my earlier fix for the removed \`--path\` flag used \`bundle config set --global path /usr/local/bundle\`, which writes to \`/home/deploy/.bundle/config\`. The production stage's Dockerfile only copies \`/app\` and \`/usr/local/bundle\` from the builder stage (not the deploy user's home directory), so that path setting never reached the production image. Bundler then couldn't find where gems were installed, hence \`Could not find rails-6.1.7.9 ... in locally installed gems\`.

Fixed by switching to \`--local\` scope, which writes to \`/app/.bundle/config\` instead — and \`/app\` **is** copied into production (\`COPY --from=builder /app .\`). Verified by building the actual \`production\` Docker target locally and confirming \`bundle exec rake -T\` loads cleanly inside it (it failed with the exact same error as the CI log before this fix, and succeeds after).

Assisted by Claude Code Sonnet 5

